PR #33, opened 2019-02-19
Color the trace output of XCVM

ir5:
This PR makes the trace output of XCVM colorful. This may be helpful for debugging XCVM.


shinh:
Thanks! I'll merge this once travis passes

PR #51, opened 2019-03-04
Enhance Pad operator

ir5:
This PR adds following features.
* Negative `pads` argument for Pad operator.
* Gradient function for Pad operator.

Some simple tests are added as well.


runtime/ops/manipulation.cc
```diff
@@ -82,14 +82,26 @@ chainerx::Array TransposeOp::RunImpl(XCVMState* st, const chainerx::Array& data)
 
 chainerx::Array PadOp::RunImpl(XCVMState* st, const chainerx::Array& data) {
     CHECK_EQ(data.ndim() * 2, pads.size());
-    chainerx::Shape shape = data.shape();
-    std::vector<chainerx::ArrayIndex> indices;
+    const auto shape = data.shape();
```
shinh: Could you use `const chainerx::Shape`, please? Overall, could you try avoiding `auto`? Copy and paste from a previous comment:

Personally, I'm not a big fan of auto. IMHO, it sometimes hurts readability significantly while it only saves you a few key strokes. I think I allow myself to use auto only for complex types. Typical examples are STL iterators, lambdas, and smart pointers.

I know this is my personal preference, but I really don't like overused `auto`..

ir5: I removed auto as possible.

PR #54, opened 2019-03-06
Computation order

ir5:
This PR adds feature to change the computation order of computation graph. Possibly recomputation can be employed.
Currently, only naive policy that performs pointless recomputation is implemented. I think we can implement more efficient recomputation strategies in future.

To run the dummy policy, flag `--computation_order dummy` must be specified when running `run_onnx`.

This PR modifies `AddNodeImpl` in `graph.h` to be a public method, which may be a little aggressive. Also, some parts are copied from `gradient.cc`, which might be subject to improve.


shinh:
LGTM, thanks!

PR #59, opened 2019-03-08
Change the second output of LRN (unit_scale) to be Opaque

ir5:
Current construction of gradient parts for LRN node is a little irregular, and it causes some error when recomputation. This PR fixes this issue, though I'm not confident if this method is the best one.


shinh:
I found #60 is a better way to fix this issue. I think this actually solves a few other problems. For example, I think we can remove "NoCtx" ops with #60 .

ir5:
Since #60 fixed this problem, I will close this PR.

PR #82, opened 2019-03-20
Gradient accumulation

ir5:


compiler/gradient_with_order.cc
```diff
@@ -190,6 +190,9 @@ void AddGradientNodesForTrainingWithOrders(Graph* graph, const std::vector<Order
                 // Copy gradients of outputs from the original
```
shinh: Could you update the comment to mention inputs are also needed to be copied?
```
                // Copy gradients of inputs/outputs from the original
                // computation node to the last forward computation.
                // Copying inputs is necessary to accumulate gradients.
```

ir5: updated.

ir5:
> add more tests

> run them both for dummy, chen, and upcoming algorithms.

I agree that these are important features that we will need to implement in future.

PR #85, opened 2019-03-22
Large testcase

ir5:
I'm adding large test cases that are generated by `onnx-chainer`. This can be used in `runtests.py` when `--use_gpu` flag is specified.
I noticed that this test fails when `two_phase_backprop` option is employed in `runtests.py`. Probably we need to investigate the reason for this.

### DIFF
--- a/scripts/gen_large_tests_oc.py
+++ b/scripts/gen_large_tests_oc.py
@@ -1,0 +1,47 @@
+#!/usr/bin/python3
+
+import chainer
+import numpy as np
+import onnx_chainer
+
+import large_models
+
+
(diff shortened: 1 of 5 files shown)


ir5:
I added VGG models in the testcase file. Unfortunately we cannot test VGG here because of dropout. Probably I can fix the problem of dropout by setting dropout-ratio=0 when generating onnx model.

ir5:
> Though it'd be even better to run this test generation by "make large_tests, I can do it later.

Currently, I did nothing for `CMakeLists.txt` and thus large tests will not be created when `make` is running. Will this cause the failure for `runtests.py`?  (I don't know why CI checks have passed...)

PR #91, opened 2019-03-25
VGG Model for large test

ir5:
I added VGG for large testcase. It seems that the gradients of VGG model incurs large numerical errors, so I set relatively high rtol/atol in `runtests.py`.

### DIFF
--- a/scripts/gen_large_tests_oc.py
+++ b/scripts/gen_large_tests_oc.py
@@ -27,26 +27,29 @@ def create_test(test_name, get_fun, dtype):
 def get_large_tests():
     tests = []
 
-    def test(name, get_fun):
+    def test(name, get_fun, kwargs=None):
+        # kwargs is used for testing
         for dtype in (np.float32, np.float64):
             output_grad = dtype == np.float64
             backprop_str = '_backprop' if output_grad else ''
             test_name = 'large_oc%s_%s_%s' % (backprop_str,
                                               name, dtype.__name__)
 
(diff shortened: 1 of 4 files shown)


scripts/gen_large_tests_oc.py
```diff
@@ -27,26 +27,29 @@ def create_test(test_name, get_fun, dtype):
...
+    test('vgg16', large_models.get_vgg16, {'rtol': 2e-2, 'atol': 2e-2})
+    test('vgg19', large_models.get_vgg19, {'rtol': 2e-2, 'atol': 2e-2})
 
     return tests
 
 
 def main():
     for test in get_large_tests():
-        create_test(*test)
+        create_test(*test[:-1])
```
shinh: Any reason not to generate the last test?

ir5: `test` here is a tuple of `(test_name, get_fun, dtype, kwargs)`. I removed the last element `kwargs` because it is not needed to create the testcase.

shinh: Ah, I see. I read the code wrongly. Then, I'd prefer something like

```
for test_name, get_fun, dtype, _ in get_large_tests():
  create_test(test_name, get_fun, dtype)
```

but it's OK as is!

ir5: Changed the code. I will merge this PR after CI check has passed.

PR #98, opened 2019-03-27
in_recomputing flag for BN

ir5:
This PR will prevent from updating batch normalization statistics more than once when recomputation is performed.
Currently, there is no test for checking the statistics. I just verified the result by manually putting `std::cout << var << std::endl;` in the intermediate of the code and seeing the values.

### DIFF
--- a/runtime/ops/normalization.cc
+++ b/runtime/ops/normalization.cc
@@ -124,6 +124,7 @@ std::tuple<chainerx::Array, XCVMOpaque*, chainerx::Array, chainerx::Array, chain
 
     PreprocessBatchNormResult result;
     if (in_recomputing) {
+        // Statistics shouldn't be updated when recomputing
         result = PreprocessBatchNorm(x, s, bias, mean.Copy(), var.Copy(), axes);
     } else {
         result = PreprocessBatchNorm(x, s, bias, mean, var, axes);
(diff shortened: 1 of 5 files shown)


runtime/ops/normalization.cc
```diff
@@ -127,7 +127,20 @@ std::tuple<chainerx::Array, XCVMOpaque*, chainerx::Array, chainerx::Array, chain
             x.device().GetBatchNormForwardBackward(result.mean, result.var, epsilon, decay, result.sorted_axis);
     const Array& gamma_reshaped = result.gamma;
     const Array& beta_reshaped = result.beta;
+
```
shinh: I'm not 100% sure, but I think we can copy `mean` and `var` before passing them to `PreprocessBatchNorm`? I.e.,
```
PreprocessBatchNormResult result;
if (in_recomputing) {
  result = PreprocessBatchNorm(x, s, bias, mean.Copy(), var.Copy(), axes);
} else {
  result = PreprocessBatchNorm(x, s, bias, mean, var, axes);
}
```
In this way, we won't need to revert the change since the inplace modification will be done for copied mean/var?

ir5: It sounds much better. I would try.

ir5: updated.

PR #100, opened 2019-03-31
Fixed Dockerfile

akmaru:
I cannot build docker image using Dockerfile.tmpl, so fixed it.
Since the following pip3 installations are lacked likely, they are added in this PR.

- onnxruntime
- pytest


shinh:
Thanks!

PR #130, opened 2019-04-07
fix: base_name was being rewritten.

Rishav1:
The if check for generator was useless as base_name was being rewritten.


durswd:
Thank you !

PR #151, opened 2019-04-10
Updated .gitignore to ignore data/

kumasento:
Hi, maybe it is better to ignore `data/` in `.gitignore`, which will be created after running `setup.sh`. Thanks


shinh:
Thanks!

Issue #160, opened 2019-04-13
Support functions and links


durswd:
Functions and layers should be supported as CH2O

## CH2O Testcase

https://github.com/pfnet-research/chainer-compiler/tree/master/ch2o/tests/node

## elichia functions

https://github.com/pfnet-research/chainer-compiler/blob/master/elichika/elichika/parser/functions_builtin.py

https://github.com/pfnet-research/chainer-compiler/blob/master/elichika/elichika/functions_buildin.py

### How to add a function

it is an example.

https://github.com/pfnet-research/chainer-compiler/pull/166

## elichika links

https://github.com/pfnet-research/chainer-compiler/blob/master/elichika/elichika/parser/links_builtin.py

https://github.com/pfnet-research/chainer-compiler/blob/master/elichika/elichika/links_builtin.py

## Rest tests
| Name | Type | Status | CH2O |
|:-----------|------------:|:------------:|:------------:|
| EmbedID.py | Links | | https://github.com/pfnet-research/chainer-compiler/blob/master/ch2o/ch2o/links.py#L455 |
| NStepBiLSTM.py | Links | | https://github.com/pfnet-research/chainer-compiler/blob/master/ch2o/ch2o/links.py#L312 |
| NStepLSTM.py | Links | | https://github.com/pfnet-research/chainer-compiler/blob/master/ch2o/ch2o/links.py#L177 |
| Roi.py | |  not pass | |
| Sum.py | Functions | | https://github.com/pfnet-research/chainer-compiler/blob/master/ch2o/ch2o/funcs.py#L483 |

Rishav1:
#205 Implemented LRN.py. Might as well add that.

durswd:
https://github.com/pfnet-research/chainer-compiler/pull/274

PR #161, opened 2019-04-14
feat: Implemented average pooling 2d.

Rishav1:
The supported args cannot be tuple right now. This is WIP. Will extend
it to support tuple arguments.

Addresses: issue #160


durswd:
Thank you. I will improve about arguments.

durswd:
I support tuple arguments

https://github.com/pfnet-research/chainer-compiler/pull/163

PR #190, opened 2019-04-23
[WIP] Support onnx chainer in python

ir5:


python/chainer_compiler.py
```diff
@@ -12,6 +12,9 @@ def _is_array(v):
 
 
 def _flatten(xs):
+    if _is_array(xs):
```
shinh: If this is really necessary, I think something is wrong.

python/chainer_compiler.py
```diff
@@ -159,18 +178,28 @@ def __init__(self, model, inputs, dump_onnx=False):
...
+                                      self.translator)
 
         graph = chainer_compiler_core.load(f.name)
         os.unlink(f.name)
 
         self.orig_output_names = graph.output_names()
 
-        fwd_graph, bwd_graph = graph.backward_to(graph.input_names())
+        # fwd_graph, bwd_graph = graph.backward_to(graph.input_names())
+        fwd_graph, bwd_graph = graph.backward_to(graph.param_names())
```
shinh: I think gradient nodes should be generated not only for params but also for inputs since chainer_compiler.compile may be called to compile only a part of a whole graph. For example, if a user compiles only ResBlock 3, gradients of inputs should be computed for backprop of ResBlock 1 and 2.

python/chainer_compiler_core.cc
```diff
@@ -95,10 +97,24 @@ std::shared_ptr<runtime::XCVM> Compile(
     return std::make_shared<runtime::XCVM>(xcvm_prog);
 }
 
+bool IsParam(Value* value) {
+    const std::string& name = value->name();
+    // the second condition is for ch2o
```
shinh: This is fine, but I'd like to see the check for `/` in Python side. Anyway, could you add "TODO(hamaji): Remove the check for '/' after deprecating CH2O" or something?

ir5:
I created this PR based on too old master repository... I'll create another PR based on newest one.

PR #196, opened 2019-04-23
Support onnx_chainer in python

ir5:


python/chainer_compiler_test.py
```diff
@@ -334,15 +357,16 @@ def test_partially_differentiable(device_name):
     model.to_device(device)
 
     expected_loss, expected_grads = _run_fwd_bwd(model, [xs, indices])
-    expected_gxs = [x.grad for x in xs]
+    # expected_gxs = [x.grad for x in xs]
```
shinh: Memo for myself: check this later

ir5:
Thank you for comments.

> If this is really necessary, I think something is wrong.

In this line, https://github.com/pfnet-research/chainer-compiler/blob/3fcc6726f02c40261b15eee0f205c37e0acd47c8/python/chainer_compiler.py#L151
the variable `tmpl` can be a `np.ndarray`. In such situation, previous implementation of `_flatten` did not work properly.

> I think gradient nodes should be generated not only for params but also for inputs since chainer_compiler.compile may be called to compile only a part of a whole graph. For example, if a user compiles only ResBlock 3, gradients of inputs should be computed for backprop of ResBlock 1 and 2.

> This is fine, but I'd like to see the check for / in Python side. Anyway, could you add "TODO(hamaji): Remove the check for '/' after deprecating CH2O" or something?

I fixed the code.

shinh:
Thanks! Re my comment on `_flatten`: you are right, sorry for my bad.

PR #205, opened 2019-04-26
feat: Passing AlexNet compilation in Elichika.

Rishav1:
Added support for Local Response Normalization and fixed an error in
AlexNet test.


scripts/elichika_tests.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, dirname, filename, fail=False):
     # of SoftmaxCrossEntropy.
     Generator('model', 'MLP', fail=True),
     # TODO(hamaji): Build and run this test.
```
shinh: Could you remove this line?

Rishav1: Sure.

shinh:
Thanks!

PR #224, opened 2019-05-09
Support Pad+Conv merging

take-cheeze:
- [x] Add Pad+Conv merging function
- [x] Add tests


compiler/merge.cc
```diff
@@ -53,6 +53,72 @@ bool MaybeMergeSplitConcat(Graph* graph, Node* node) {
...
+    }
+    Node* conv = pad->outputs()[0]->users()[0];
+    if (conv->op_type() != Node::kConv) {
+        return false;
+    }
+
+    // replace node
+    GraphBuilder gb(graph, "MergePadConv", pad->input(0));
+    std::vector<Value*> new_in = pad->inputs();
+    new_in.insert(new_in.end(), conv->inputs().begin(), conv->inputs().end());
```
take-cheeze: Needs to filter `Pad` node from `conv->inputs()`

compiler/merge.cc
```diff
@@ -53,6 +53,71 @@ bool MaybeMergeSplitConcat(Graph* graph, Node* node) {
...
 }
 
+bool MaybeMergePadConv(Graph* graph, Node* pad) {
+    if (pad->value() != 0.0 || pad->mode() != "constant") {
+        return false;
+    }
+
+    auto const& pads = pad->pads();
+
+    // range check
```
shinh: Could you make this comment a complete sentence? For example, "Padding for non-spatial dims cannot be merged."

I forgot to mention, but I'm trying to follow ChainerX's coding guideline (https://github.com/chainer/chainer/blob/master/docs/source/chainerx/contribution.rst#coding-standards), which is based on Google's. 
 Google's says comments should be complete sentences (https://google.github.io/styleguide/cppguide.html#Punctuation,_Spelling_and_Grammar).

compiler/merge.cc
```diff
@@ -53,6 +53,71 @@ bool MaybeMergeSplitConcat(Graph* graph, Node* node) {
...
+    if (pads.size() < 4) {
+        return false;
+    }
+
+    // pads should apply to feature dims
```
shinh: If you use the comment I suggested (i.e., "Padding for non-spatial dims cannot be merged."), I think we can remove this comment (and remove blank line in line 67)?

compiler/merge.cc
```diff
@@ -53,6 +53,71 @@ bool MaybeMergeSplitConcat(Graph* graph, Node* node) {
...
+    if (pads[0] != 0 || pads[1] != 0 || pads[pads.size() / 2] != 0 || pads[pads.size() / 2 + 1] != 0) {
+        return false;
+    }
+
+    // pads of begin and end must be same
```
shinh: Ditto. "Pads of begin and end must be same."

compiler/merge.cc
```diff
@@ -53,6 +53,71 @@ bool MaybeMergeSplitConcat(Graph* graph, Node* node) {
...
+    for (auto i = 2; i < pads.size() / 2; ++i) {
+        if (pads[i] != pads[pads.size() / 2 + i]) {
+            return false;
+        }
+    }
+
+    // pads must be greater or equal to 0
```
shinh: ditto

compiler/merge.cc
```diff
@@ -53,6 +53,71 @@ bool MaybeMergeSplitConcat(Graph* graph, Node* node) {
...
+    if (std::any_of(pads.begin(), pads.end(), [](auto p) { return p < 0; })) {
+        return false;
+    }
+
+    // connected node must be Conv
```
shinh: ditto

compiler/merge.cc
```diff
@@ -53,6 +53,71 @@ bool MaybeMergeSplitConcat(Graph* graph, Node* node) {
     return true;
 }
 
+bool MaybeMergePadConv(Graph* graph, Node* pad) {
+    if (pad->value() != 0.0 || pad->mode() != "constant") {
+        return false;
+    }
+
+    auto const& pads = pad->pads();
```
shinh: `const std::vector<int64_t>&` please. This is my personal preference, but I'm not a big fun of `auto` for simple types.

Copy and paste from a previous comment in another review:

Personally, I'm not a big fan of auto. IMHO, it sometimes hurts readability significantly while it only saves you a few key strokes. I think I allow myself to use auto only for complex types. Typical examples are STL iterators, lambdas, and smart pointers.

compiler/merge.cc
```diff
@@ -53,6 +53,71 @@ bool MaybeMergeSplitConcat(Graph* graph, Node* node) {
...
+
+    // pads of begin and end must be same
+    for (auto i = 2; i < pads.size() / 2; ++i) {
+        if (pads[i] != pads[pads.size() / 2 + i]) {
+            return false;
+        }
+    }
+
+    // pads must be greater or equal to 0
+    if (std::any_of(pads.begin(), pads.end(), [](auto p) { return p < 0; })) {
```
shinh: [optional] This is fine as is, but I feel simple `for` loop is shorter and easy to read.

```
for (int64_t p : pads) {
  if (p < 0) {
    return false;
  }
}
```

compiler/merge.cc
```diff
@@ -66,6 +131,10 @@ void MergeOperations(Graph* graph) {
             if (node->op_type() == Node::kSplit) {
                 replaced |= MaybeMergeSplitConcat(graph, node);
             }
+
+            if (node->op_type() == Node::kPad) {
```
shinh: `else if`?

compiler/merge.cc
```diff
@@ -53,6 +53,71 @@ bool MaybeMergeSplitConcat(Graph* graph, Node* node) {
...
+
+    // replace node
+    GraphBuilder gb(graph, "MergePadConv", pad->input(0));
+    std::vector<Value*> new_in = pad->inputs();
+    std::copy(conv->inputs().begin() + 1, conv->inputs().end(), std::back_inserter(new_in));
+    Node* n = gb.MOp(Node::kConv, new_in, conv->outputs());
+    n->set_dilations(conv->dilations());
+    n->set_group(conv->group());
+    n->set_kernel_shape(conv->kernel_shape());
+    n->set_strides(conv->strides());
```
shinh: Please add `n->set_auto_pad(conv->auto_pad())` just in case?

compiler/merge.cc
take-cheeze: OK. I'll follow it!

shinh:
Still looks great. Merged, thanks!

PR #228, opened 2019-05-13
Replace internal of Tensor class with chainerx::Array

take-cheeze:


compiler/shape_evaluator.cc
```diff
@@ -78,10 +79,11 @@ void DoEvaluateShape(Node* node) {
             const Type& type = input->type();
             int64_t nbytes = type.GetNBytes();
             CHECK_LT(0, nbytes);
-            Tensor::UniqueData data(std::malloc(nbytes), &std::free);
+            std::unique_ptr<void, decltype(&std::free)> data(std::malloc(nbytes), &std::free);
```
shinh: Maybe we can use `chainerx::Zeros`?

compiler/tensor.cc
```diff
@@ -83,16 +86,13 @@ void DumpDataToRepeated(const Tensor& t, ::google::protobuf::RepeatedField<To>*
...
 
-}  // namespace
-
-Tensor::Tensor(const onnx::TensorProto& xtensor)
-    : dims_(xtensor.dims().begin(), xtensor.dims().end()),
-      dtype_(xtensor.data_type()),
-      data_(nullptr, &std::free),
-      name_(xtensor.name()),
-      doc_string_(xtensor.doc_string()) {
+chainerx::Array TensorProtoToArray(onnx::TensorProto const& xtensor) {
```
shinh: Please revive initializations for `name_` and `doc_string_`. My bad this was not caught by a test...

compiler/tensor.cc
```diff
@@ -101,83 +101,90 @@ Tensor::Tensor(const onnx::TensorProto& xtensor)
         CHECK_EQ(0, xtensor.double_data_size());
         CHECK_EQ(0, xtensor.uint64_data_size());
 
-        switch (dtype_) {
+        int64_t const num_elements = shape.GetTotalSize();
+        switch (dtype) {
```
shinh: Do we still need this switch? I think we can just use `MakeHostArray(dtype, shape, xtensor.raw_data())`?

compiler/evaluator.cc
```diff
@@ -25,35 +25,8 @@ namespace chainer_compiler {
...
-
 Tensor* ArrayToTensor(const std::string& name, const chainerx::Array& a) {
-    Tensor::UniqueData data(std::malloc(a.GetNBytes()), &std::free);
-    memcpy(data.get(), a.Copy().ToNative().raw_data(), a.GetNBytes());
-    std::vector<int64_t> dims{a.shape().begin(), a.shape().end()};
-    return new Tensor(name, GetDtype(a), dims, data.release());
+    return new Tensor(name, a);
 }
 
 }  // namespace
```
shinh: I think we can simplify tensor->array conversion thanks to the change for Tensor. https://github.com/pfnet-research/chainer-compiler/pull/228/files#diff-3bb5e0736d88411209b08e07e5eec2eeR77 could be just `state.SetArray(t->array());`. I think you need to add the new accessor function to `Tensor`. (or `Tensor::chx()` for consistency with Dtype?)

compiler/tensor.cc
take-cheeze: This just splits the creation of `chainerx::Array` so constructor is still there.

compiler/evaluator.cc
```diff
@@ -117,7 +78,8 @@ void Eval(
 
         switch (var->kind()) {
             case runtime::XCVMVar::Kind::kArray: {
-                outputs->emplace_back(new EvaluatedValue(ArrayToTensor(name, state.GetArray(output_id))));
+                // TODO(take-cheeze): Avoid copy if possible.
```
shinh: Thanks for adding this TODO!

PR #233, opened 2019-05-14
Add convolution operator support in FLOPs estimater

take-cheeze:
Most of the operator is moved to #266 
Operators going to support with tests:
- [x] ConvTranspose
- [x] ChainerConvGradWeight
- [x] Gemm(adding tests)


take-cheeze:
Looks there is conflict so I'll fix it.

compiler/flops.cc
```diff
@@ -32,22 +32,66 @@ int64_t OutputSize(const Node& node) {
...
-    int64_t flops = node.output(0)->type().NumElements();
+    int64_t const out_size = node.output(0)->type().NumElements();
+    int64_t flops = out_size;
     flops *= node.input(0)->type().dims()[node.trans_a() ? 0 : 1];
+    if (node.alpha() != 1.0) {
+        flops += out_size;
+    }
+    if (node.beta() != 0.0) {
+        flops += out_size * node.input(2)->type().dims()[node.trans_a() ? 0 : 1];
+        if (node.beta() != 1.0) {
```
shinh: I guess we don't need this extra cost for `0 < beta < 1` since it could be done by FMA?

compiler/flops.cc
```diff
@@ -32,22 +32,66 @@ int64_t OutputSize(const Node& node) {
...
 int64_t CalculateFlopsOfGemm(const Node& node) {
-    int64_t flops = node.output(0)->type().NumElements();
+    int64_t const out_size = node.output(0)->type().NumElements();
+    int64_t flops = out_size;
     flops *= node.input(0)->type().dims()[node.trans_a() ? 0 : 1];
+    if (node.alpha() != 1.0) {
+        flops += out_size;
+    }
+    if (node.beta() != 0.0) {
+        flops += out_size * node.input(2)->type().dims()[node.trans_a() ? 0 : 1];
```
shinh: Not sure, but I think `flops += out_size` is correct. For A (NxM), B (MxK), and C(NxK), Computation cost of AxB is O(NMK) but the cost of X+C is O(NxK).

compiler/flops.cc
```diff
@@ -32,22 +32,66 @@ int64_t OutputSize(const Node& node) {
...
 
 int64_t CalculateFlopsOfConv(const Node& node) {
-    int64_t bsize = node.input(0)->type().dims()[0];
-    int64_t ichan = node.input(0)->type().dims()[1];
-    int64_t kw = node.input(1)->type().dims()[2];
-    int64_t kh = node.input(1)->type().dims()[3];
-    int64_t ochan = node.output(0)->type().dims()[1];
-    int64_t ow = node.output(0)->type().dims()[2];
-    int64_t oh = node.output(0)->type().dims()[3];
+    Type const& x = node.input(0)->type();
```
shinh: Thanks for these renaming :)

compiler/flops.cc
```diff
@@ -32,22 +32,66 @@ int64_t OutputSize(const Node& node) {
...
+    Type const& w = node.input(1)->type();
+    Type const& y = node.output(0)->type();
+    int64_t bsize = x.dims()[0];
+    int64_t ichan = x.dims()[1];
+    int64_t kw = w.dims()[2];
```
shinh: This is my fault, but it seems all of `w` and `h` are swapped :( The third dimension is `h` and the forth is `w`. This won't change the calculation though..

compiler/flops.cc
take-cheeze: So can I expect it to be fused in the future?
(Current implementation isn't fused:
https://github.com/pfnet-research/chainer-compiler/blob/db3e2ec3aedf8654b7f730c11b7ce8c934322a14/runtime/ops/math.cc#L162

compiler/flops.cc
take-cheeze: Seems that's right. I'll fix it.

compiler/flops.cc
shinh: Good point. Yeah, I feel the current implementation is just inadequate. I think we should just call sgemm function in OpenBLAS or cuBLAS, which would have the sane implementation, I believe. I think it'd be better to show flops of sane implementation since this number will be used to estimate theoritical limit latency.

PR #243, opened 2019-05-16
Backprop two phase with order

ir5:
Implemented two_phase + computation_order option for backprop.

I tested only in CPU environment, though I believe that this should work in GPU environment as well.


compiler/gradient_with_order.cc
```diff
@@ -243,4 +245,192 @@ void AddGradientNodesForTrainingWithOrders(Graph* graph, const std::vector<Order
...
+        gbf.Op(Node::kIdentity, {value}, o);
+
+        Value* i = bwd_graph->AddInputValue(name, value->type());
+
+        // Update the staged value to the retained one
+        p.second = i;
+    }
+}
+
+void AddGradientNodesForTrainingWithOrders(Graph* fwd_graph, Graph* bwd_graph, const std::vector<Order>& orders) {
```
shinh: I think this function is very similar to the single graph mode. Cannot we just call this version from the other version by `AddGradientNodesForTrainingWithOrders(graph, graph, orders)`? I think, it should just work with a few `fwd_graph == bwd_graph` check in theory?

compiler/computation_order/policy_dummy.cc
```diff
@@ -31,4 +31,32 @@ std::vector<Order> DummyPolicy(const Graph& graph) {
...
+    std::vector<Order> orders;
+
+    std::set<Value*> params;
+    for (Value* value : graph.input_values()) {
+        params.insert(value);
+    }
+
+    auto nodes = graph.GetTopologicallySortedNodes();
+    for (auto node : nodes) {
+        orders.emplace_back(Order::kComputeForward, node, nullptr);
```
shinh: Is this a valid scheduling for all computation graphs? For a graph like
```
v1 = n1(v0)
v2 = n2(v1)
v3 = n3(v1)
v4 = n4(v2, v3)
```
I think the `orders` will be
```
Compute n1
Compute n2
Forget v1
Compute n3
```
at this point, `v1` is not ready? If my concern is right, I think you can use `graph->GetNecessaryNodesAndInputCounts` and do something like
```
zzz = graph->GetNecessaryNodesAndInputCounts()
for (node : nodes) {
  if (!params.count(node) && --zzz[node] == 0)
}
```
I haven't checked this work, though

ir5: No, dummy2 will output some error for some family of computational graphs.
My intention of adding this policy was to check if retaining intermediate values is working well in two_phase+computation_order mode for simple case.
I do not want to use this policy in practical case.

shinh: Got it. Could you add `CHECK_EQ(1, value->users().size()) << "Graph with branches is not compatible with this computation order policy";` then?

compiler/gradient_with_order.cc
ir5: I'm not sure if it is possible, but I will try. At least, I think we need to avoid calling `AddRetainedParts` when `fwd_graph == bwd_graph`. I don't know if I'm missing other things.

compiler/computation_order/policy_dummy.cc
ir5: Sure, I would.

compiler/gradient_with_order.cc
ir5: I merged the two versions of `AddGradientNodesForTrainingWithOrders`. It seems working.

compiler/computation_order/policy_dummy.cc
ir5: done

compiler/gradient_with_order.cc
```diff
@@ -132,18 +159,41 @@ void AddGradientNodesForTrainingWithOrders(Graph* graph, const std::vector<Order
...
+        if (fwd_graph != bwd_graph && current_graph == fwd_graph) {
+            bool output_all_staged = true;
+            for (Value* output : fwd_graph->output_values()) {
+                if (!staged.count(output)) output_all_staged = false;
+            }
+            if (output_all_staged) {
+                current_graph = bwd_graph;
+                {
+                    ScheduleAddedScope schedule_scope(fwd_graph, schedule_node);
+                    {
```
shinh: No need for the extra nest?

compiler/gradient_with_order.cc
```diff
@@ -202,10 +257,9 @@ void AddGradientNodesForTrainingWithOrders(Graph* graph, const std::vector<Order
                     }
                 }
 
-                ScheduleAddedScope schedule_scope(graph, schedule_node);
-                if (!AddGradientForNode(graph, graph, node, nullptr)) {
+                ScheduleAddedScope schedule_scope(bwd_graph, schedule_node);
+                if (!AddGradientForNode(bwd_graph, bwd_graph, node, nullptr)) {  // NOTE: first argument may be fwd_graph?
```
shinh: Yeah, this the first argument should be `fwd_graph`.

compiler/gradient_with_order.cc
shinh: Thanks!

ir5:
```
build/tools/run_onnx --test out/backprop_test_pc_for_float32 --compiler_log --backprop --computation_order=dummy
```

Thank you for information.
Unfortunately, it seems that this test fails even in current master branch. I don't know when this test broke... I want to fix this in future. Probably we need to develop sophisticated way of testing recomputation methods.

compiler/gradient_with_order.cc
ir5: Without this nest, I could not compile the code. 

```
[ 86%] Building CXX object compiler/CMakeFiles/chainer_compiler_compiler.dir/gradient_with_order.cc.o
/home/mkusumoto/chainer-compiler/compiler/gradient_with_order.cc: In function ‘void chainer_compiler::AddGradientNodesForTrainingWithOrders(chainer_compiler::Graph*, chainer_compiler::Graph*, const std::vector<chainer_compiler::Order>&)’:
/home/mkusumoto/chainer-compiler/compiler/gradient_with_order.cc:181:59: error: redeclaration of ‘chainer_compiler::{anonymous}::ScheduleAddedScope schedule_scope’
                         ScheduleAddedScope schedule_scope(bwd_graph, schedule_node);
                                                           ^
/home/mkusumoto/chainer-compiler/compiler/gradient_with_order.cc:180:40: note: ‘chainer_compiler::{anonymous}::ScheduleAddedScope schedule_scope’ previously declared here
                     ScheduleAddedScope schedule_scope(fwd_graph, schedule_node);

```

shinh: Renaming variables should work?
```
ScheduleAddedScope fwd_schedule_scope(fwd_graph, schedule_node);
ScheduleAddedScope bwd_schedule_scope(bwd_graph, schedule_node);
```

ir5: Yes, you are true. (sorry for silly mistake...) Fixed.

compiler/gradient_with_order.cc
ir5: It seems that, if we change the first argument to `fwd_graph`, `runtests.py` yields more filed tests than current one.

```
# When first argument = bwd_graph:

$ python ./scripts/runtests.py --computation_order=dummy
...
65/976 tests failed! (see out/failed_tests.log)

# When first argument = fwd_graph:

$ python ./scripts/runtests.py --computation_order=dummy
...
75/976 tests failed! (see out/failed_tests.log)
```

Honestly, I do not totally figure out what is correct. I may seek what is really correct in future.

PR #253, opened 2019-05-17
Add Where node support

take-cheeze:


runtime/xcvm_defs.py
```diff
@@ -373,6 +373,8 @@ def sigil(typ):
       String('code'), Int('fusion_id')],
      [ArrayList('outputs')]),
 
+    ('Where', [Array('condition'), Array('x'), Array('y')], [Array('output')]),
```
shinh: s/condition/cond/ for consistency with .cc file.

take-cheeze:
Thanks for the review! Should be fixed now.

PR #263, opened 2019-05-20
Skip test case with computation_order if dyamic shape is used

ir5:
Several tests will be skipped in runtests.py with computation_order flag.

### DIFF
--- a/compiler/gradient_with_order.cc
+++ b/compiler/gradient_with_order.cc
@@ -132,7 +132,20 @@ void AddRetainedParts(Graph* fwd_graph, Graph* bwd_graph, std::map<Value*, Value
     }
 }
 
-void AddGradientNodesForTrainingWithOrders(Graph* fwd_graph, Graph* bwd_graph, const std::vector<Order>& orders) {
+bool IsComputationOrderSupported(const Graph& graph) {
+    for (const auto& value_ptr : graph.all_values()) {
+        if (value_ptr.get()->type().GetNBytes() < 0) {
+            return false;
+        }
+    }
+    return true;
+}
+
+bool AddGradientNodesForTrainingWithOrders(Graph* fwd_graph, Graph* bwd_graph, const std::vector<Order>& orders) {
+    if (!IsComputationOrderSupported(*fwd_graph) || !IsComputationOrderSupported(*bwd_graph)) {
+        return false;
+    }
+
     // A map from the original value to the staged value, possibly recomputed.
     std::map<Value*, Value*> staged;
     for (Value* value : fwd_graph->input_values()) {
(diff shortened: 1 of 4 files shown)


compiler/gradient_with_order.cc
```diff
@@ -132,7 +132,20 @@ void AddRetainedParts(Graph* fwd_graph, Graph* bwd_graph, std::map<Value*, Value
     }
 }
 
-void AddGradientNodesForTrainingWithOrders(Graph* fwd_graph, Graph* bwd_graph, const std::vector<Order>& orders) {
+bool IsComputationOrderSupported(const Graph& graph) {
+    for (const auto& value_ptr : graph.all_values()) {
```
shinh: `for (const Value* value : graph.GetNecessaryValeus())` would be better since we don't care if unused value has dynamic shape?

compiler/passes.cc
```diff
@@ -111,7 +111,11 @@ void RunDefaultPasses(Graph* graph, bool gen_backprop) {
             // specified computation order
             skip_scheduling = true;
             auto orders = GetComputationOrder(*graph, g_computation_order);
-            AddGradientNodesForTrainingWithOrders(graph, orders);
+            if (!AddGradientNodesForTrainingWithOrders(graph, orders)) {
+                std::cerr << "Computation order is not supported in this graph.\n";
```
shinh: Doesn't `CHECK(false) << ...` won't work for you? Silently exiting with zero wouldn't be a great idea since it may get people overlook errors.

tools/run_onnx.cc
```diff
@@ -240,7 +240,11 @@ class ModelRunner {
                 GenerateGradientNodes(model->mutable_graph(), backprop_model.mutable_graph());
             } else {
                 auto orders = GetComputationOrder(model->graph(), g_computation_order);
-                AddGradientNodesForTrainingWithOrders(model->mutable_graph(), backprop_model.mutable_graph(), orders);
+                if (!AddGradientNodesForTrainingWithOrders(model->mutable_graph(), backprop_model.mutable_graph(), orders)) {
+                    LOG() << "Computation order is not supported in this graph." << std::endl;
```
shinh: ditto?

compiler/gradient_with_order.cc
ir5: fixed.

compiler/passes.cc
ir5: It's a little difficult point. I wanted to suppress errors caused by this part when I run `python scripts/runtests.py --computation_order=dummy`. But as you say, we may easily overlook errors with this code.
Adding a "skipped" status in runtests.py may be convenient in this case, but I'm not sure if it is the best.

shinh: OK. I guess we can figure out a better solution, but let's go ahead with this. Thanks for your explanation!

PR #264, opened 2019-05-20
Run computation_order tests on CI

shinh:


shinh:
Nah, probably, this isn't a great idea. I think the list should be a whitelist, not a blacklist, since newly added tests will fail.

shinh:
Closing in favor of https://github.com/pfnet-research/chainer-compiler/pull/265

PR #270, opened 2019-05-22
Add and improve operators(Abs/Power/Sign)

take-cheeze:
To follow latest ChainerX updates.

- [x] Support [`Sign`](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Sign)
- [x] Use `chainerx::Absolute`
- [x] Use `chainerx::Power`
- [x] Resolve LRN TODOs
- [x] Update chainer version

### DIFF
--- a/runtime/ops/math.cc
+++ b/runtime/ops/math.cc
@@ -15,9 +15,7 @@ namespace runtime {
 namespace {
 
 chainerx::Array Pow(chainerx::Array a, chainerx::Array b) {
-    return chainerx::Exp(chainerx::Log(a) * b);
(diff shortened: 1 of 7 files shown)


runtime/ops/math.cc
```diff
@@ -16,6 +16,8 @@ namespace {
 
 chainerx::Array Pow(chainerx::Array a, chainerx::Array b) {
     return chainerx::Exp(chainerx::Log(a) * b);
+    // TODO(take-cheeze): Update ChainerX to use element wise power from chainer/chainer#6496.
+    // return chainerx::Power(a, b);
```
shinh: Could you resolve a couple of TODOs for LRN in normalization.cc?

PR #275, opened 2019-05-23
Export models using unified memory

ir5:
To export models in a memory-efficient manner, I wrote a code to use unified memory.
Further memory reduction such as exporting the model with batchsize=1 is probably possible. I start from easy point.


python/chainer_compiler.py
```diff
@@ -162,6 +167,48 @@ def backward(self, indexes, flat_gys):
         return gxs
 
 
+class use_unified_memory:
```
shinh: Could you use `contextlib.contextmanager`? https://docs.python.org/2/library/contextlib.html

python/chainer_compiler.py
```diff
@@ -6,6 +6,11 @@
 import ch2o
 import chainer_compiler_core
 
+try:
+    import cupy
+except Exception:
```
shinh: `ImportError` please

python/chainer_compiler.py
```diff
@@ -162,6 +167,48 @@ def backward(self, indexes, flat_gys):
...
+
+def _run_translator(translator, mc, inputs):
+    if translator == 'ch2o':
+        xmodel = ch2o.compile_model(mc, inputs)
+        f = tempfile.NamedTemporaryFile(delete=False)
+        f.write(xmodel.SerializeToString())
+        f.close()
+        del xmodel
+    elif translator == 'onnx_chainer':
+
```
shinh: An unnecessary blank line?

python/chainer_compiler.py
```diff
@@ -180,23 +227,9 @@ def __init__(self, model, inputs, translator='ch2o', dump_onnx=False,
             self.compile(inputs)
 
     def compile(self, inputs):
```
shinh: As chatted locally, I think it's better to make this feature available only when the user explicitly specify a parameter. Otherwise, end users may be confused by `to_gpu`-ed process consuming a lot of host memory.

python/chainer_compiler.py
ir5: fixed. (`yield None` is OK in this case?)

python/chainer_compiler.py
ir5: fixed

python/chainer_compiler.py
ir5: fixed

python/chainer_compiler.py
ir5: Added use_unified_memory option.

python/chainer_compiler.py
```diff
@@ -162,16 +168,52 @@ def backward(self, indexes, flat_gys):
...
+
+    return graph
+
+
 class CompiledModel(chainer.Chain):
 
     def __init__(self, model, inputs, translator='ch2o', dump_onnx=False,
-                 computation_order=None):
+                 computation_order=None,
+                 use_unified_memory=False):
```
shinh: As chatted, it would be better to change the name of this argument to `allocators` or something which will be a tuple of (runtime allocator, export time allocator).

python/chainer_compiler.py
```diff
@@ -162,16 +168,52 @@ def backward(self, indexes, flat_gys):
         return gxs
 
 
+@contextmanager
+def enable_unified_memory(use):
+    if use:
+        cupy.cuda.set_allocator(cupy.cuda.memory.malloc_managed)
+    try:
+        yield None
```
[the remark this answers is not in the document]
ir5: Thanks, I removed the `with` block implementation because it was simpler.

python/chainer_compiler.py
ir5: I added arguments of export_allocator and runtime_allocator.

python/chainer_compiler.py
```diff
@@ -180,23 +211,11 @@ def __init__(self, model, inputs, translator='ch2o', dump_onnx=False,
...
-            f.close()
-        else:
-            raise NotImplementedError('Unsupported translator:',
-                                      self.translator)
-
-        graph = chainer_compiler_core.load(f.name)
-        os.unlink(f.name)
+        if self.export_allocator is not None:
+            cupy.cuda.set_allocator(self.export_allocator)
+        graph = _run_translator(self.translator, self.mc, inputs)
```
shinh:
```
try:
  graph = ...
finally:
  if self.runtime_allocator is not None:
    ...
```

just in case?

python/chainer_compiler.py
```diff
@@ -162,16 +167,42 @@ def backward(self, indexes, flat_gys):
...
+                 export_allocator=None,
+                 runtime_allocator=None):
         super(CompiledModel, self).__init__()
         with self.init_scope():
             self.mc = model
         self.translator = translator
         self.dump_onnx = dump_onnx
         self.computation_order = computation_order
+        self.export_allocator = export_allocator
+        self.runtime_allocator = runtime_allocator
```
shinh: Could you add something like
```
if export_allocator is not None and runtime_allocator is None:
  raise ValueError('`runtime_allocator` must be set when `export_allocator` is set')
```
I think this combination would indicate the caller does not understand these parameters.

python/chainer_compiler.py
ir5: Thanks, I modified.

python/chainer_compiler.py
ir5: Added.

PR #283, opened 2019-05-28
Use std::function instead of template

shinh:


shinh:
ChainerCI, test this please

PR #289, opened 2019-05-28
Add Shape and Scalar type to ChxVMVar

### LINKED ISSUE #18: Add scalar/shape types to XCVMVar
As of Jan 2019, all tensor values are encoded by chainerx::Array, which is inefficient. It'd be better to introduce kScalar and kShape to handle these values without chainerx::Array. Also, we probably want to

* introduce std::variant
* change XCVMState::variables_ from unique_ptr to non-unique_ptr

take-cheeze:
- closes #18
- Only adds enum and getter to ChxVMVar
- Adding abseil for `variant` type since currently chainer-compiler uses C++14


take-cheeze:
Making the output of Shape operator `chainerx::Shape` seems to be a good start point to use this

PR #293, opened 2019-05-29
Enhance test cases for computation order

ir5:
This PR should be merged after #291.

This PR enhances the test of computation order by introducing manually added computation order (CustomPolicy).
This PR also fixes a bug of computation order detected by these tests.

### DIFF
--- a/scripts/runtests.py
+++ b/scripts/runtests.py
@@ -651,7 +651,7 @@ if not args.all:
 
 def _start_output(msg):
     if sys.stdout.isatty():
-        sys.stdout.write('\r' + ' ' * 120 + '\r' + msg)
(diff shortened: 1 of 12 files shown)


scripts/runtests.py
```diff
@@ -631,7 +651,7 @@
 
 def _start_output(msg):
     if sys.stdout.isatty():
-        sys.stdout.write('\r' + ' ' * 78 + '\r' + msg)
+        sys.stdout.write('\r' + ' ' * 120 + '\r' + msg)
```
shinh: Hmm I want to keep the number of spaces as is. I think we need to abbreviate the status message when a line is too long. If you want, I can do it by myself.

ir5: I just reverted this part. I left the abbreviation of the status message as future work.

PR #297, opened 2019-05-30
Register custom ONNX operators as dynamic operator

### LINKED ISSUE #220: Debug build fails in ONNX
```
$ cmake .. -DCHAINERX_BUILD_PYTHON=OFF -DCHAINERX_BUILD_CUDA=OFF -G Ninja -DCMAKE_BUILD_TYPE=Debug
$ ninja
$ build-dbg/tools/run_onnx --test third_party/onnx/onnx/backend/test/data/node/test_abs
Initializing ChainerX...
Loading model...
Loading data...
Found 1 test cases
Constructing model...
terminate called after throwing an instance of 'chainer_compiler_onnx::assert_error'
  what():  ../third_party/onnx/onnx/defs/schema.cc:824: SchemasRegisterer: Assertion `dbg_registered_schema_count == DbgOperatorSetTracker::Instance().GetCount()` failed: 236 schema were exposed from operator sets and automatically placed into the static registry.  248 were expected based on calls to registration macros. Operator set functions may need to be updated.
```

Probably, the right solution is to create a custom ONNX op registry instead of adding custom ops to opset 9, but I'm going to disable assertion for now.

take-cheeze:
According to this comment:
https://github.com/onnx/onnx/blob/5972eed96752551a0ea8b932904fedd568c48d3b/onnx/defs/schema.h#L853-L872

closes #220

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,9 +109,6 @@ add_subdirectory(third_party/onnx)
 # It seems `onnx` target compiles onnx*.pb.cc. As this will let other
 # files produce warnings, we suppress all warnings from ONNX by -w.
 target_compile_definitions(onnx PRIVATE ONNX_API=)
-# TODO(hamaji): Remove this workaround after fixing the following issue:
-# https://github.com/pfnet-research/chainer-compiler/issues/220
-target_compile_definitions(onnx PRIVATE NDEBUG=1)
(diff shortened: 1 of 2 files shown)


shinh:
Oops I commented on issue instead of this PR. Heh, I didn't expect the solution is this easy. Thanks for the nice finding!

PR #305, opened 2019-05-31
Specialize GeItem from shape variable

take-cheeze:
relates to: #289

When `kShape` tagged `ChxVMVar` is referenced as `chainerx::Array` it converts `chainerx::Shape` internally.
Though I'm not sure whether some member variable needs to be qualified `mutable`.

Only supports `GenericGetItem` op in this PR.

### DIFF
--- a/runtime/chxvm_defs.py
+++ b/runtime/chxvm_defs.py
@@ -216,7 +216,7 @@ XC_OPS = [
     ('Elu', [Array('x'), Float('alpha')], ['y']),
     ('Floor', [Array('x')], ['y']),
     ('Ceil', [Array('x')], ['y']),
-    ('Shape', [Array('data')], ['shape']),
+    ('Shape', [Array('data')], ['shape'], False),
     ('Size', [Array('data')], ['size']),
     ('Reshape', [Array('data'), Array('shape')], ['reshaped']),
     ('Expand', [Array('input'), Array('shape')], ['output']),
--- a/runtime/chxvm_defs.py
+++ b/runtime/chxvm_defs.py
@@ -89,6 +89,8 @@ class ValueInfo(_ValueInfo):
             return 'array_list'
         elif self.typ == OPAQUE:
             return 'opaque'
+        elif self.typ == SHAPE:
+            return 'shape'
         else:
             raise RuntimeError('Unknown type: %s' % self.typ)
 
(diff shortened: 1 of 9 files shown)


runtime/chxvm_var.cc
```diff
@@ -1,4 +1,5 @@
 #include "runtime/chxvm_var.h"
+#include "runtime/chainerx_util.h"
```
shinh: Sort alphabetically?

runtime/chxvm_var.h
```diff
@@ -49,6 +49,8 @@ class ChxVMVar {
     explicit ChxVMVar(chainerx::Array array);
     // Takes the ownership of `opaque`.
     explicit ChxVMVar(ChxVMOpaque* opaque);
+    explicit ChxVMVar(chainerx::Shape opaque);
```
shinh: s/opaque/shape/

runtime/chxvm_var.h
```diff
@@ -49,6 +49,8 @@ class ChxVMVar {
...
+    explicit ChxVMVar(chainerx::Scalar opaque);
```
shinh: ditto

runtime/chxvm_defs.py
```diff
@@ -216,7 +216,7 @@ def sigil(typ):
     ('Elu', [Array('x'), Float('alpha')], ['y']),
     ('Floor', [Array('x')], ['y']),
     ('Ceil', [Array('x')], ['y']),
-    ('Shape', [Array('data')], ['shape']),
+    ('Shape', [Array('data')], ['shape'], False),
```
shinh: This is fine for now. However, I think we should introduce `Shape` type to this IDL-ish script. The Shape op should look like `('Shape', [Array('data')], [Shape('shape')])` and Reshape will look like `('Reshape', [Array('data'), Shape('shape')], ['reshaped'])`.

runtime/ops/manipulation.cc
```diff
@@ -8,8 +8,8 @@
 namespace chainer_compiler {
 namespace runtime {
 
-chainerx::Array ShapeOp::RunImpl(ChxVMState* st, const chainerx::Array& data) {
-    return ShapeToArray(data.shape());
+void ShapeOp::RunImpl(ChxVMState* st) {
```
shinh: I want the code generator to generate the prototype like `chainerx::Shape ShapeOp::RunImpl(ChxVMState* st, const chainerx::Array& data);` and returned shape will be automatically `SetShape`-ed.

runtime/chxvm_var.cc
take-cheeze: I don't know why but this seems to be done by clang-format.

take-cheeze:
I've tried and failed the new return and argument type when writing #289 but I've succeeded this time!

runtime/chxvm_var.cc
shinh: :thinking: Then, it's fine. Thanks for your explanation!

shinh: Ah, now I can see the reason. This was because this .cc file was chxvm_var.cc. Google's style guide says foo.h should come first in foo.cc:

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

shinh: The follow up will be done here: https://github.com/pfnet-research/chainer-compiler/pull/306/commits/96f18f5e4345bdfc57e0e2f625f9afbba85033a9

PR #306, opened 2019-05-31
Minor follow up for #305

shinh:


shinh:
ChainerCI, test this please

PR #311, opened 2019-06-01
Support for continue (including nesting).

Rishav1:
Also contains
- test for break (Canonicalizer for that is WIP).
- implementation for simplifying `ast.UnaryOp(ast.USub, ast.Num(1))` to `ast.Num(-1)` 

I am working on test structure for Canonicalizer.

### DIFF
--- a/elichika/elichika/parser/canonicalizer.py
+++ b/elichika/elichika/parser/canonicalizer.py
@@ -22,12 +22,11 @@ class Canonicalizer(NodeTransformer):
             return node
 
     def visit_For(self, node):
-        self.for_continue_stack.append([])
-        node = self.generic_visit(node)
+        modified_node = self.generic_visit(node)
         continue_flags = self.for_continue_stack.pop()
         for flag in continue_flags:
             node.body.insert(0, ast.Assign(targets=[ast.Name(id=flag, ctx=ast.Store())], value=ast.NameConstant(value=False)))
-        return node
+        return modified_node
 
     def generic_visit(self, node):
         if isinstance(node, ast.stmt):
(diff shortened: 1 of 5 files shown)


elichika/elichika/parser/canonicalizer.py
```diff
@@ -0,0 +1,65 @@
...
+    def generic_visit(self, node):
+        if isinstance(node, ast.stmt):
+            if len(self.for_continue_stack) > 0 and len(self.for_continue_stack[-1]) > 1:
+                bool_values = []
+                for flag in self.for_continue_stack[-1]:
+                    bool_values.append(ast.UnaryOp(op=ast.Not(), operand=ast.Name(id=flag, ctx=ast.Load())))
+                if isinstance(node, ast.For):
+                    self.for_continue_stack.append([])
+                node = super().generic_visit(node)
+                replacement = ast.If(test=ast.BoolOp(op=ast.And, values=bool_values), body=[node], orelse=[])
```
shinh: Maybe you could do something like
```
if len(bool_values) == 1:
  cond = bool_values[0]
else:
  cond = ast.BoolOp(...)
```
?

elichika/tests/syntax/Continue.py
```diff
@@ -0,0 +1,58 @@
...
+            if i == y - 1:
+                continue
+            for j in range(x):
+                if i == y:
+                    if j == y:
+                        continue
+                ret = ret + j
+        return ret
+
+
```
shinh: It seems `len(self.for_continue_stack[-1]) > 1` is not tested? I tried to add a test case for this case but it failed due to a bug in elichika. I think the `assert canonicalize(to_ast(source)) == to_ast(target)` style test you are adding will cover the case, but it will be helpful for us to add a commented-out test like (please just use TODO(hamaji))

https://github.com/pfnet-research/chainer-compiler/blob/master/elichika/tests/syntax/Alias.py#L54

so that we won't forget the issue.

Of course, fixing the bug itself is also welcomed :)

elichika/elichika/parser/functions.py
```diff
@@ -366,10 +367,10 @@ def __init__(self, func):
         self.lineno = sourcelines[1]
 
         code = utils.clip_head(inspect.getsource(func))
-
+        processed_ast = canonicalizer.Canonicalizer().visit(ast.parse(code))
```
shinh: I'm not sure, but maybe better to keep `gast.ast_to_gast` before the canonicalization? Otherwise, the canonicalizer may need to take care of multiple python versions? It seems gast also has NodeTransformer so I'm guessing we can use it.

PR #316, opened 2019-06-03
Support scalar type in ChxVM

take-cheeze:
Also adds optional scalar type that corresponds to optional array type.

- [x] Fix `ToString` stuff
- [x] Support `Float/IntScalarConstant`
- [x] Type `Shape` in some operators


runtime/chxvm_var.cc
```diff
@@ -41,9 +43,33 @@ ChxVMVar::ChxVMVar(chainerx::Shape shape) : kind_(Kind::kShape), val_(shape) {
...
+                    break;
+                }
+                case chainerx::DtypeKind::kInt: {
+                    int64_t i(s);
+                    val_ = runtime::MakeArray(chainerx::Dtype::kInt64, {}, &i);
+                    break;
+                }
+                case chainerx::DtypeKind::kFloat: {
+                    double f(s);
+                    val_ = runtime::MakeArray(chainerx::Dtype::kFloat64, {}, &f);
```
shinh: I think this can cause unexpected dtype promotions. Maybe scalar type in `VarInternalType` must be a pair of `chainerx::Scalar` and `chainerx::Dtype`?

take-cheeze: Yes. And in many cases it affects the type inferences too.
So I'm defining new type to represent scalar type.

take-cheeze:
Now this should work. The equality check with `ChxVMVar::Kind` is kinda difficult.

shinh:
ChainerCI, test this please

runtime/chxvm_var.cc
```diff
@@ -59,9 +75,14 @@ ChxVMOpaque* ChxVMVar::GetOpaque() const {
     return absl::get<std::shared_ptr<ChxVMOpaque>>(val_).get();
 }
 
-const chainerx::Scalar& ChxVMVar::GetScalar() const {
+const StrictScalar& ChxVMVar::GetScalar() const {
+    if (kind_ == Kind::kArray) {
+        const chainerx::Array& ary = absl::get<chainerx::Array>(val_);
+        val_ = StrictScalar(ary.dtype(), chainerx::AsScalar(ary));
```
shinh: I think host bit has been gone here.
```
        val_ = StrictScalar(ary.dtype(), chainerx::AsScalar(ary), IsNativeDevice(&ary.device()));
```

shinh:
ChainerCI, test this please

PR #318, opened 2019-06-03
Installation via pip

ir5:
Update: Now ready for review

This PR will enable to install chianer-compiler via pip.
TODO:

- [x] Fix hard-coded paths
- [x] Fix examples
- ~Update requirements to install the appropriate version of chainer~
- [x] Catch up the latest master branch

One concern:
Current code under `python` and `example` directories is assuming that chainer_compiler is installed via `setup.py`. However, CI will run without the installation via `setup.py`.
So, in test code, I put a little complicated code to modify path configuration. We may improve these parts in the future.
Also, we will have to  add documentation about the installation via pip.


setup.py
```diff
@@ -0,0 +1,72 @@
...
+            '-DCHAINER_COMPILER_ENABLE_CUDA=ON',
+            '-DCHAINERX_BUILD_CUDA=ON',
+            '-DCUDA_TOOLKIT_ROOT_DIR=' + cuda_path,
+            '-DCHAINER_COMPILER_ENABLE_PYTHON=ON',
+            '-DPYTHON_EXECUTABLE=' + python3_executable,
+            '-DCHAINERX_BUILD_PYTHON=ON',
+            '-DCUDNN_ROOT_DIR=' + cudnn_root_dir,
+        ]
+        build_args = [
+            '--', '-j4',
```
shinh: I think this can kill machines with small RAMs. Shouldn't we just rely on `MAKEFLAGS` set outside this script?

ir5: I agree with you. I would like to add `MAKEFLAGS` as an optional flag.

setup.py
```diff
@@ -0,0 +1,69 @@
...
+        if cudnn_root_dir is None:
+            raise RuntimeError('CUDNN_ROOT_DIR must be set.')
+
+        cmake_args = [
+            '-DCMAKE_LIBRARY_OUTPUT_DIRECTORY=' + extdir,
+            '-DCHAINER_COMPILER_ENABLE_CUDA=ON',
+            '-DCHAINERX_BUILD_CUDA=ON',
+            '-DCUDA_TOOLKIT_ROOT_DIR=' + cuda_path,
+            '-DCHAINER_COMPILER_ENABLE_PYTHON=ON',
+            '-DPYTHON_EXECUTABLE=' + python3_executable,
```
disktnk: `sys.executable` returns python path

python/chainer_compiler/chainer_compiler.py
```diff
@@ -169,6 +179,7 @@ def backward(self, indexes, flat_gys):
 
 def _run_translator(translator, mc, inputs):
     if translator == 'ch2o':
+        import ch2o
```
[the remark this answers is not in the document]
ir5: Yes, we should consider how to import ch2o and elichika. I want to leave this task as future work, though.

setup.py
ir5: Thank you, I modified the code.

shinh:
LGTM, please merge this if this PR is ready.

ir5:
Thank you, I would merge.

PR #319, opened 2019-06-03
Fix sigil of C++ side

take-cheeze:
Since it seems python code is correct.


take-cheeze:
I can remove the `sigil` method but the duplication would still remain like this:
https://gist.github.com/take-cheeze/580f917c7090b6d4495b2a54ca90bf2a

shinh:
Thanks for the info! I'm feeling `colored_name` should be moved to C++ side as well, but I'm not sure if it's a great idea. Let's get this PR merged anyway!

PR #326, opened 2019-06-04
Use `variant::index()` instead of `kind_` member in `ChxVMVar`

take-cheeze:


take-cheeze:
ChainerCI, test this please

shinh:
Still LGTM

PR #338, opened 2019-06-06
Use PREBUILT option in setup.py

ir5:


setup.py
```diff
@@ -29,6 +29,8 @@ def build_cmake(self, ext):
             os.path.dirname(self.get_ext_fullpath(ext.name)))
 
         cuda_path = os.getenv('CUDA_PATH', '/usr/local/cuda')
+        chainerx_preduilt_path = os.getenv(
```
disktnk: supporting to check prebuild binary looks good!

how about get directory path dynamically when building environment has chainerx module?


first, 

```
import os
import chainerx
chainerx_prebuilt_path = os.path.dirname(chainerx.__file__)
```

When causing `ImportError`, next checking environment variable like this PR

disktnk: preduilt -> prebuilt ?

ir5: Thank you for advice, I modified the code as such.

PR #339, opened 2019-06-07
Use ngraph/tvm even without --fuse_operations

shinh:


shinh:
ChainerCI, test this please

PR #340, opened 2019-06-07
[WIP] Modify import path of ch2o/elichika and the python binding

ir5:
Work in progress.

This PR is to put the packages of `ch2o` and `elichika` under `chainer_compiler` directory. This way, we will be able to import `ch2o` by `import chainer_compiler.ch2o`. Both packages should be installed via `setup.py`.

I modified various path configurations.


shinh:
ChainerCI, test this please

PR #351, opened 2019-06-11
Remove PYTHONPATH on pip install CI

disktnk:
to check pip install more stable


disktnk:
ChainerCI, test this please

PR #365, opened 2019-06-14
Generate builtin json getter with python script

take-cheeze:


configs/CMakeLists.txt
```diff
@@ -1,16 +1,17 @@
 get_filename_component(CHAINER_COMPILER_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR} PATH)
 
 set(jsons
-  chxvm.json
-  chxvm_test.json
-  dldt.json
-  ngraph.json
-  nvrtc.json
-  tvm.json
+  chxvm
```
shinh: I think I may put a new target before this element, so.. (please check the next comment)

configs/gen_load_json_from_name.py
```diff
@@ -0,0 +1,30 @@
...
+
+import sys
+
+
+def main(argv):
+    print('#include <configs/json_repository.h>')
+    print('#include <common/log.h>')
+    print('namespace chainer_compiler {')
+    print('namespace builtin_configs {')
+    for j in argv[1:]:
```
shinh: Could you move 'chxvm' to the front?

```
>>> i = argv.index('chxvm')
>>> [argv[i]] + argv[:i] + argv[i+1:]
```

configs/CMakeLists.txt
take-cheeze: Good point. I'm feared about it too.

configs/gen_load_json_from_name.py
```diff
@@ -4,6 +4,9 @@
 
 
 def main(argv):
+    argv.remove('chxvm')
```
shinh: Ah, this is easier than what I suggested :)

PR #370, opened 2019-06-14
Support break

Rishav1:
- Added `keepgoing` variable during canonicalization that represents the exit-condition for break.
- Added support for `BoolOp` multiary operators And and Or (WIP).
- Currently passing 2/3 `Break.py` tests.

### DIFF
--- a/chainer_compiler/elichika/onnx_converters.py
+++ b/chainer_compiler/elichika/onnx_converters.py
@@ -478,12 +478,13 @@ def convert_node_multiary_op(onnx_graph, node: 'nodes.NodeMultiaryOp'):
 
     if node.multiaryop == nodes.MultiaryOpType.And:
         op = 'And'
+        temp_prev = ONNXValue(onnx_graph, np.array(True, dtype=np.bool), [node, '/True'], is_constant=True)
     elif node.multiaryop == nodes.MultiaryOpType.Or:
         op = 'Or'
+        temp_prev = ONNXValue(onnx_graph, np.array(False, dtype=np.bool), [node, '/False'], is_constant=True)
     else:
         assert(False)
 
-    temp_prev = ONNXValue(onnx_graph, np.array(True, dtype=np.bool), [node, '/True'], is_constant=True)
     for idx, value_ in enumerate(node.values_list):
         temp = ONNXValue(onnx_graph, np.array(True).dtype, [node, '/temp%d' % idx])
         onnx_graph.add_node(
--- a/chainer_compiler/elichika/onnx_converters.py
+++ b/chainer_compiler/elichika/onnx_converters.py
@@ -474,6 +474,32 @@ def convert_node_call(onnx_graph, node: 'nodes.NodeCall'):
         chainer_l_converter[type(original_inst)](onnx_graph, node)
 
 
+def convert_node_multiary_op(onnx_graph, node: 'nodes.NodeMultiaryOp'):
+
+    if node.multiaryop == nodes.MultiaryOpType.And:
+        op = 'And'
+    elif node.multiaryop == nodes.MultiaryOpType.Or:
+        op = 'Or'
+    else:
+        assert(False)
+
+    temp_prev = ONNXValue(onnx_graph, np.array(True, dtype=np.bool), [node, '/True'], is_constant=True)
+    for idx, value_ in enumerate(node.values_list):
+        temp = ONNXValue(onnx_graph, np.array(True, dtype=np.bool), [node, '/temp%d' % idx])
+        onnx_graph.add_node(
+            op,
+            [temp_prev.name, value2onnx_parameter[value_].onnx_name],
+            [temp.name],
+            str(node.lineprop))
+        temp_prev = temp
+
+    onnx_graph.add_node(
+        "Identity",
+        [temp_prev.name],
+        [value2onnx_parameter[node.outputs[0]].onnx_name],
+        str(node.lineprop))
+
+
 def convert_node_unary_op(onnx_graph, node: 'nodes.NodeUnaryOp'):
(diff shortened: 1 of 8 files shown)


chainer_compiler/elichika/onnx_converters.py
```diff
@@ -474,6 +474,32 @@ def convert_node_call(onnx_graph, node: 'nodes.NodeCall'):
...
+    if node.multiaryop == nodes.MultiaryOpType.And:
+        op = 'And'
+    elif node.multiaryop == nodes.MultiaryOpType.Or:
+        op = 'Or'
+    else:
+        assert(False)
+
+    temp_prev = ONNXValue(onnx_graph, np.array(True, dtype=np.bool), [node, '/True'], is_constant=True)
+    for idx, value_ in enumerate(node.values_list):
+        temp = ONNXValue(onnx_graph, np.array(True, dtype=np.bool), [node, '/temp%d' % idx])
```
Rishav1: Not sure why but the compiler throws an error that the `temp0` and 'temp1' is the output of more than one node nodes. Any clue why this is happening? The testcase is [this](https://github.com/pfnet-research/chainer-compiler/blob/b064ec2eb8f4ebe077b2f8dec02822a38963cde6/testcases/elichika_tests/syntax/Break.py#L30-L39). The convert_node_multiary_op is hit when processing the canonicalized syntax--
```
if continued_1 or breaked_0:
    ...
```

```
*** Testing out/elichika_tmp/tmp_break_continue_mixed ***
build/tools/run_onnx --test out/elichika_tmp/tmp_break_continue_mixed
Initializing ChainerX...
Loading model...
Loading data...
Found 1 test cases
Constructing model...
ERROR: A value is output more than once: `MultiaryOp(L.-1,MultiaryOpType.And)MultiaryOp(L.-1,MultiaryOpType.And)/temp0` creator: And(MultiaryOp(L.-1,MultiaryOpType.And)MultiaryOp(L.-1,MultiaryOpType.And)/True, continued_1_st_st_st_L.-1) -> (MultiaryOp(L.-1,MultiaryOpType.And)MultiaryOp(L.-1,MultiaryOpType.And)/temp0)
ERROR: A value is output more than once: `MultiaryOp(L.-1,MultiaryOpType.And)MultiaryOp(L.-1,MultiaryOpType.And)/temp1` creator: And(MultiaryOp(L.-1,MultiaryOpType.And)MultiaryOp(L.-1,MultiaryOpType.And)/temp0, breaked_0_st_st_L.-1_2) -> (MultiaryOp(L.-1,MultiaryOpType.And)MultiaryOp(L.-1,MultiaryOpType.And)/temp1)
Failed graph is stored in /tmp/chainer_compiler_failure.onnx
Check `false' failed! in CheckSanity at /home/rchouras/Workplace/chainer-compiler/compiler/graph.cc:376: Sanity check (SSA) failed
Traceback (most recent call last):
  File "scripts/elichikacheck.py", line 44, in <module>
    main()
  File "scripts/elichikacheck.py", line 41, in main
    subprocess.check_call(args)
  File "/usr/lib/python3.6/subprocess.py", line 291, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['build/tools/run_onnx', '--test', 'out/elichika_tmp/tmp_break_continue_mixed']' died with <Signals.SIGABRT: 6>.
```

shinh: It seems `temp%d` appear as outputs twice (`Constant` and `And`). Looking at code around this function, it seemed we should pass dtypes instead of actual arrays to create non-constant `ONNXValue`. The following change worked for me:

```
diff --git a/chainer_compiler/elichika/onnx_converters.py b/chainer_compiler/elichika/onnx_converters.py
index 3365bec..d726d14 100644
--- a/chainer_compiler/elichika/onnx_converters.py
+++ b/chainer_compiler/elichika/onnx_converters.py
@@ -485,7 +485,7 @@ def convert_node_multiary_op(onnx_graph, node: 'nodes.NodeMultiaryOp'):
 
     temp_prev = ONNXValue(onnx_graph, np.array(True, dtype=np.bool), [node, '/True'], is_constant=True)
     for idx, value_ in enumerate(node.values_list):
-        temp = ONNXValue(onnx_graph, np.array(True, dtype=np.bool), [node, '/temp%d' % idx])
+        temp = ONNXValue(onnx_graph, np.array(True).dtype, [node, '/temp%d' % idx], is_constant=False)
         onnx_graph.add_node(
             op,
             [temp_prev.name, value2onnx_parameter[value_].onnx_name],
@@ -666,7 +666,7 @@ class ONNXValue:
                 self.tensor = onnx_graph.new_tensor_with_np(
                     self.np_value, self.name)
 
-        elif(any_value == np.float32 or any_value == np.float64 or any_value == np.int32 or any_value == np.int64):
+        elif(any_value == np.float32 or any_value == np.float64 or any_value == np.int32 or any_value == np.int64 or any_value == np.bool):
             self.name = generate_name()
             self.tensor = self.onnx_graph.new_empty_tensor(
                 ['TODO'], any_value, self.name)

```

@durswd could you verify the above change? Also, I found the constructor of `ONNXValue` is error prone. I think it'd be better to define a helper function like `create_onnx_constant` and `create_onnx_tensor` or something to make intention of callers clearer.

Rishav1: Should I go ahead and make the suggested change?

Rishav1: Also, I guess I should add tests for BoolOp syntax.

shinh: Yeah, please go ahead. Thanks!

PR #375, opened 2019-06-18
Replace setup.sh by downloading in cmake build

take-cheeze:


runtime/ops/connection.cc
```diff
@@ -27,6 +27,22 @@ chainerx::Array ConvOp::RunImpl(
     Int64StackVector comp_strides = ComplementStride(strides, x);
     Int64StackVector comp_pads = ComplementPad(pads, x);
 
+    if (auto_pad == "SAME_UPPER") {
+        for (size_t i = 0; i < comp_pads.size(); ++i) {
+            int64_t& pad = comp_pads[i];
+            const int64_t in_dim = x.shape()[i];
```
take-cheeze: Seems I've forgot `+ 2` here

shinh: Wow, thanks for the catch!

take-cheeze:
ChainerCI, test this please!

PR #376, opened 2019-06-18
Fix naming correspondence when BN is used

ir5:
Work in progress.

This PR fixes the naming correspondence of `bn.avg_mean` and `bn.avg_var`. This PR should be merged after https://github.com/chainer/onnx-chainer/issues/181 is fixed.


ir5:
Update: Now ready for review

chainer_compiler/chainer_compiler.py
```diff
@@ -203,6 +203,34 @@ def _run_translator(translator, mc, inputs):
     return graph
 
 
+def _resolve_name_correspondence(mc, params, name, param_values):
+    for running_mean_name in ('avg_mean', 'running_mean'):
```
shinh: Any reason we need to support both `avg_mean` and `running_mean`? Also, could you put a link to onnx-chainer to mention this code assumes the naming convention of onnx-chainer?

chainer_compiler/chainer_compiler.py
```diff
@@ -203,6 +203,34 @@ def _run_translator(translator, mc, inputs):
     return graph
 
 
+def _resolve_name_correspondence(mc, params, name, param_values):
```
shinh: How about `_resolve_name_correspondence_bn`? This logic is very specific for BN.

chainer_compiler/chainer_compiler.py
```diff
@@ -203,6 +203,34 @@ def _run_translator(translator, mc, inputs):
...
+    for running_mean_name in ('avg_mean', 'running_mean'):
+        if not name.endswith(running_mean_name):
+            continue
+        parent = name[:-len(running_mean_name)]
+        mean = params[parent + 'beta']
+        for link in mc.links():
+            if not hasattr(link, 'avg_mean'):
+                continue
+            if any(id(p) == id(mean) for p in link.params()):
```
shinh: I think `links.params` iterates the parameters recursively. If a user defines a chain with `avg_mean` defined, this function may misbehave.

chainer_compiler/chainer_compiler.py
```diff
@@ -203,6 +203,34 @@ def _run_translator(translator, mc, inputs):
     return graph
 
 
+def _resolve_name_correspondence(mc, params, name, param_values):
+    for running_mean_name in ('avg_mean', 'running_mean'):
+        if not name.endswith(running_mean_name):
+            continue
+        parent = name[:-len(running_mean_name)]
+        mean = params[parent + 'beta']
+        for link in mc.links():
```
shinh: I think iterating over `mc.links` should be done around line 300, before `for name in self.param_names:`.

chainer_compiler/chainer_compiler.py
```diff
@@ -265,23 +293,35 @@ def compile(self, inputs):
...
+        fwd_chxvm_vars = fwd_graph.params()
+        for name in self.param_names:
+            if name in params:
+                self.param_values.append(params[name])
+            else:
+                out = _resolve_name_correspondence(
+                    self.mc, params, name, self.param_values)
+                if not out:
+                    # Retrieve the initial value from ONNX initializer
+                    if name in fwd_chxvm_vars:
```
shinh: Do we really need this fallback path?

chainer_compiler/chainer_compiler.py
```diff
@@ -203,6 +203,34 @@ def _run_translator(translator, mc, inputs):
...
 
 
+def _resolve_name_correspondence(mc, params, name, param_values):
+    for running_mean_name in ('avg_mean', 'running_mean'):
+        if not name.endswith(running_mean_name):
+            continue
+        parent = name[:-len(running_mean_name)]
+        mean = params[parent + 'beta']
+        for link in mc.links():
+            if not hasattr(link, 'avg_mean'):
```
shinh: If possible, it'd be better to check `isinstance(link, chainer.links.BatchNormalization)`?

chainer_compiler/chainer_compiler.py
```diff
@@ -265,23 +293,35 @@ def compile(self, inputs):
 
         self.compiled = True
 
+        params = dict(self.mc.namedparams())
+        if self.translator == 'onnx_chainer':
+            params = {'param' + key.replace('/', '_'): value for key, value
+                      in params.items()}
```
shinh: I think we can update `params` by iterating `namedlinks` here. Maybe something like

```
for link_name, link in self.mc.namedlinks().items():
  if not isinstance(link, BatchNormalization):
    continue
  link_name = 'param' + link_name.replace('/', '_')
  for avg_name in ['avg_mean', 'avg_var']:
    key = link_name + '_' + avg_name
    assert key not in params
    params[key'] = getattr(link, avg_name)
```

chainer_compiler/chainer_compiler.py
```diff
@@ -203,6 +203,34 @@ def _run_translator(translator, mc, inputs):
...
+        if not name.endswith(running_mean_name):
+            continue
+        parent = name[:-len(running_mean_name)]
+        mean = params[parent + 'beta']
+        for link in mc.links():
+            if not hasattr(link, 'avg_mean'):
+                continue
+            if any(id(p) == id(mean) for p in link.params()):
+                param_values.append(link.avg_mean)
+                return True
```
shinh: I think it's better to return `link.avg_mean` instead of True / False and not to modify `param_values` in this function, so that caller will look more consistent.

chainer_compiler/chainer_compiler.py
ir5: This approach looks concise. I modified the large part of the code to this kind of approach.

chainer_compiler/chainer_compiler.py
ir5: Without this part, I couldn't handle the input value with name such as `MulConstant_0_param_value`.  If we improve onnx_chainer further, we may be able to get rid of this part...?

chainer_compiler/chainer_compiler.py
```diff
@@ -265,23 +265,43 @@ def compile(self, inputs):
...
+        if self.translator == 'ch2o':
+            convert_rule = lambda key: key  # noqa
+        elif self.translator == 'onnx_chainer':
+            convert_rule = lambda key: 'param' + key.replace('/', '_')  # noqa
+
+        params = {convert_rule(key): value for key, value
+                  in dict(self.mc.namedparams()).items()}
+
+        # Since avg_mean and avg_var in BatchNormalization are not parameters
+        # in chianer link, we need an additional handling.
```
shinh:
```suggestion
        # in chainer link, we need an additional handling.
```

chainer_compiler/chainer_compiler.py
```diff
@@ -265,23 +265,43 @@ def compile(self, inputs):
...
+        for link_name, link in dict(self.mc.namedlinks()).items():
```
shinh:
```suggestion
        for link_name, link in self.mc.namedlinks():
```
would work?

chainer_compiler/chainer_compiler.py
```diff
@@ -265,23 +265,43 @@ def compile(self, inputs):
...
+                assert key not in params
+                params[key] = getattr(link, avg_name)
+
+        self.param_values = []
+        fwd_chxvm_vars = fwd_graph.params()
+        for name in self.param_names:
+            if name in params:
+                self.param_values.append(params[name])
+            elif name in fwd_chxvm_vars:
+                # Retrieve the initial value from ONNX initializer
```
shinh: Could you add
```
TODO(hamaji): Emit `Constant` in onnx-chainer so we will not need this branch.
```

chainer_compiler/chainer_compiler.py
shinh: Thanks for your explanation! Now I see the reason. Yeah, I guess onnx-chainer should emit `Constant` instead of initializer for such values. Could you add a TODO I'll comment?

chainer_compiler/chainer_compiler.py
ir5: Thank you. Added.

chainer_compiler/chainer_compiler.py
ir5: Fixed.

chainer_compiler/chainer_compiler.py
ir5: Sorry for silly typo... Fixed.

Issue #328, opened 2019-06-04
Support `None` type


take-cheeze:
Currently they are treated as `False` in ch2o:
https://github.com/pfnet-research/chainer-compiler/blob/22981097b33fcddda586003b169644bba5a8c05b/ch2o/ch2o/utils.py#L112

How to check:
```
diff --git a/ch2o/tests/syntax/Cmp.py b/ch2o/tests/syntax/Cmp.py
index fa8913b..e1ab3fc 100644
--- a/ch2o/tests/syntax/Cmp.py
+++ b/ch2o/tests/syntax/Cmp.py
@@ -67,6 +67,7 @@ if __name__ == '__main__':
         for x, y in [(None, None),
                      (42, None),
                      (True, None),
+                     (False, None),
                      (True, False),
                      (True, True),
                      (False, False),
@@ -75,4 +76,3 @@ if __name__ == '__main__':
                      (np.array(42), np.array(42))]:
             ch2o.generate_testcase(cls(), [x, y],
                                    subname='%s_%s_%s' % (x, name, y))
-
```

`$ python3 scripts/ch2ocheck.py ch2o/tests/syntax/Cmp.py`

Thoughts to implement:
- Make current `Null` kind of `ChxVMVar` undefined
- Mark a special opaque kind `None`

PR #383, opened 2019-06-19
[WIP] Support None type

take-cheeze:
#328


take-cheeze:
Seems there is problem representing `None` in initializers of nodes.
`TensorProto` can only handle tensor type(of course), though there is a candidate type to use like `None`:
- `STRING` type
- `UNDEFINED` type

Though either of the type may cause some confusion or unexpected errors....

take-cheeze:
Seems difficult for now so doing this later

PR #385, opened 2019-06-19
Support QuantizeLinear op

take-cheeze:
https://github.com/onnx/onnx/blob/master/docs/Operators.md#QuantizeLinear


take-cheeze:
ChainerCI, test this please!

PR #389, opened 2019-06-19
Support return

Rishav1:
- Updated Canonicalizer to move return statement to end. Multiple returns are serialized into one.
- Added tests.


Rishav1:
> Thanks! I haven't look into the change yet, but could you check the failure on travis? https://travis-ci.org/pfnet-research/chainer-compiler/builds/547811468?utm_source=github_status&utm_medium=notification

The canonicalizer was not handling the case when user defined `__init__` function had no explicit return statement. Fixed that.

testcases/elichika_tests/canonicalizer/Return_test.py
```diff
@@ -0,0 +1,265 @@
...
+
+        value = 0
+        for a in range(10):
+            for b in range(10):
+                value += func(a, b)
+        """)
+        target_code = utils.clip_head("""
+        def func(a, b):
+            returned_value = None
+            returned_1 = False
```
shinh: I think we don't need multiple `returned_xx` in a function. Could you modify the code for simplicity?

testcases/elichika_tests/canonicalizer/Return_test.py
```diff
@@ -0,0 +1,265 @@
...
+from chainer_compiler.elichika.parser import canonicalizer
+from chainer_compiler.elichika.parser import utils
+from chainer_compiler.elichika.testtools import compare_ast, assert_semantically_equals
+
+
+class Return(unittest.TestCase):
+    def setUp(self):
+        self.canonicalizer = canonicalizer.Canonicalizer(use_illegal_identifier=False)
+
+    def test_return(self):
```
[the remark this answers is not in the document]
Rishav1: The canonicalizer modified it to match:
```python
    def test_return_noop(self):
        orig_code = utils.clip_head("""
        def func(a, b):
            return a + b

        value = func(3, 39)
        """)
        target_code = utils.clip_head("""
        def func(a, b):
            # initializing return flags and return value assignment variable.
            returned_value = None
            returned_0 = False
            # replacing return with boolean flag and assignment. 
            returned_0 = True
            returned_value = a + b
            # returning the assigned value.
            return returned_value

        value = func(3, 39)
        """)
        orig_ast = gast.ast_to_gast(ast.parse(orig_code))
        target_ast = gast.ast_to_gast(ast.parse(target_code))
        converted_ast = self.canonicalizer.visit(orig_ast)

        assert_semantically_equals(orig_code, target_code, ['value'])
        assert compare_ast(converted_ast, target_ast)
        assert compare_ast(target_ast, converted_ast)
```

testcases/elichika_tests/canonicalizer/Return_test.py
Rishav1: The boolean `returned_1` flag is used to break a loop or when return statement is hit or to skip subsequent statements. The value to be returned is stored in `returned_value`, since different return statements might try to return different values.

I can try combining both the flag and the value variable, by inferring the value of boolean flag based on any change in `returned_value` variable, but it can cause ambiguity when there is a return statement with no return value (i.e. it doesn't change the value of the `returned_value` variable).

Is there an alternative way you have in mind that can work with just a single `returned_xx` variable?

shinh: Sorry, I'm OK with having both `returned_flag` and `returned_value_key`. I was wondering if we can combine `returned_0` and `returned_1` here.

testcases/elichika_tests/canonicalizer/Return_test.py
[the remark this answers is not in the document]
Rishav1: It actually helps in handling the following case:
```python
orig_code = utils.clip_head("""
        def func(a, b):
            if a > b:
                 return a - b
            else:
                 return b - a

        value = func(3, 39)
        """)
```
In this case, the canonicalizer replaces all the returns with assignment and there is a singular `return returned_value` statement in the last. The variable `returned_value` would store either `a-b` or `b-a` based on the branch encountered. Because of the recursive implementation, even if there is just one return statement, it is replaced by an assignment, and a new return statement is added in the end.

I can add a special edge case to not do this when there is a singular return statement, but do you think it is a good idea?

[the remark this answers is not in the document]
shinh: Ah, it sounds like the current behavior is reasonable. No, I don't think we need to handle the single return case specially. Thanks for you explanation. Could you just add `test_return_noop` case with the expected `target_code` with the current canonicalizer?

testcases/elichika_tests/canonicalizer/Return_test.py
Rishav1: Yeah, I guess that can be done. Actually, maybe this can be done for countinued_xx and breaked_xx flags as well.

Rishav1: Alright. Would make the change and see if any edge case arises.

testcases/elichika_tests/canonicalizer/Return_test.py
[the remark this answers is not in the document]
Rishav1: Yes, Sure. Will do that.

testcases/elichika_tests/canonicalizer/Return_test.py
shinh: Yeah, I was also wondering if we can reduce the number of breaked flags a bit. Thanks for fixing both of them!

PR #393, opened 2019-06-20
Start supporting quantized operator with simplifier

take-cheeze:
- [x] [QLinearConv](https://github.com/onnx/onnx/blob/master/docs/Operators.md#qlinearconv)
- [x] [QLinearMatMul](https://github.com/onnx/onnx/blob/master/docs/Operators.md#qlinearmatmul)
- [x] [MatMulInteger](https://github.com/onnx/onnx/blob/master/docs/Operators.md#MatMulInteger)
- [x] [ConvInteger](https://github.com/onnx/onnx/blob/master/docs/Operators.md#ConvInteger)


take-cheeze:
ChainerCI, test this please!

take-cheeze:
ChainerCI, test this please!

take-cheeze:
ChainerCI, test this please!

runtime/chainerx_util.h
```diff
@@ -43,5 +43,14 @@ bool IsFloat(chainerx::Dtype dtype);
 
 void BlitArray(const chainerx::Array& src, const chainerx::Array& dst);
 
+chainerx::Array MatMul(const chainerx::Array& a, const chainerx::Array& b);
```
shinh: Could you rename this to something like BatchMatmul? When ChainerX adds MatMul, I think we will silently start using chainerx::MatMul instead of this function due to ADL (which I hate, btw), and we may forget to remove this function or worse overlook incompatibility between these two functions

runtime/chainerx_util.h
```diff
@@ -43,5 +43,14 @@ bool IsFloat(chainerx::Dtype dtype);
...
+chainerx::Array Conv(
```
shinh: Again, it'd be safe to intentionally avoid the same name with ChainerX's. `Convolution`?

scripts/runtests.py
```diff
@@ -193,10 +193,9 @@
     TestCase(NODE_TEST, 'test_cast_FLOAT_to_DOUBLE'),
     TestCase(NODE_TEST, 'test_cast_FLOAT_to_FLOAT16'),
 
-    # TODO(ChainerX): Support non-2D dot.
```
shinh: Thanks for removing this TODO. It exists since the first week of this project..

PR #398, opened 2019-06-24
Fail when _chainer_compiler_core.so is missing

shinh:


shinh:
ChainerCI, test this please

PR #397, opened 2019-06-21
Remove test_bn test

shinh:
This partially reverts #376

PR #399, opened 2019-06-24
Re-enable some GPU tests

shinh:
Only test_bn has an issue. #397 disabled too much tests.


shinh:
ChainerCI, test this please

PR #400, opened 2019-06-24
Add run_onnx_dldt.py

disktnk:
- this util script does not use chainer-compiler module.


disktnk:
I confirmed the script run on

- CPU, fp32
- GPU, fp32
- GPU, fp16

with ResNet50

then remove "[WIP]"

shinh:
Thanks!!

PR #403, opened 2019-06-25
Workaround for string type in verbose mode

take-cheeze:


compiler/dtype.cc
```diff
@@ -175,6 +175,9 @@ int Dtype::SizeOf() const {
             return 4;
         case kFloat64:
             return 8;
+        case kString:
+            // TODO(take-cheeze): Make size accurate
```
shinh: FYI, I guess it was more appropriate to have `Type::Kind::kString` instead of here?

take-cheeze: Seems better. Though ONNX doesn't have string kind in value so may need some trick

PR #405, opened 2019-06-25
Start supporting `print` in elichika

take-cheeze:
- [x] Support `print` builtin function in elichika
- [x] Make nodes without outputs necessary(seems current way isn't working since those node can't be tracked from `output_values`)
- [x] Add some tests
- [x] More string support in ChxVM(though still experimental)


take-cheeze:
ChainerCI, test this please!

PR #409, opened 2019-06-26
Introduce memory efficient export

ir5:
This PR remove the restriction that the batch-size is consistent during the training.
By this way, we can export the model with batch-size one to reduce memory consumption.

Some concern:
* I made some `CHECK_EQ` to be checked only when options.check_type is true. Is this OK?
* Also, I introduced `init_allocator` function. But I'm not sure if this is really useful.


ir5:
It turned out that export with batch-size=1 is not useful. I will close this PR...

PR #414, opened 2019-06-27
Bump up chainer version to v7.0.0b1

shinh:
This branch should be ABI compatible with ChainerX v7.0.0b1


shinh:
ChainerCI, test this please

PR #419, opened 2019-06-28
Return elapsed seconds from dldt's run

shinh:


shinh:
ChainerCI test this please

PR #425, opened 2019-06-28
Set skip_scheduling from the call of python

ir5:
For compatibility with `run_onnx`


shinh:
ChainerCI, test this please

PR #428, opened 2019-06-29
Remove reduce sum to

shinh:
when shape is not changed.


shinh:
ChainerCI, test this please

PR #436, opened 2019-06-30
Supporting Dict & With syntax + interface for elichika builtin flags

Rishav1:
- [x] Update `VEvalOption` class.
- [x] Add user side support for dummy veval option flags.
- [x] Add Support for `with` syntax. (User defined managers need additional support for `__enter__` and `__exit__` calls)
- [x] Add support for 'Dict' syntax(1/2). (Dictionary declaration, setter and getter)
- [x] Add support for 'Dict' syntax(2/2). (Dictionary functions like items, keys, ~~values~~)
- [x] Add support for `ignore_branch` flag.
- [x] Add support for subscript on self. (Like `self['lstm1']`)
- [x] Add support for `for_unroll` flag.
- [x] Add constant string propagation for % operator.
- [x] Add tests for Dictionary support.
- [x] Add tests for existing `for` unroll. (When the iterator list consists of only constants)
- [x] Add tests for `ignore_branch` flag. ~~(Add tests for testing mismatch in output)~~
- [x] Add tests for `for_unroll` flag.

### DIFF
--- a/chainer_compiler/elichika/onnx_converters.py
+++ b/chainer_compiler/elichika/onnx_converters.py
@@ -317,7 +317,9 @@ def convert_node_bin_op(onnx_graph, node: 'nodes.NodeBinOp'):
         seq_right = right.create_sequence()
         onnx_graph.add_node(binops[node.binop], [seq_left, seq_right], [
                             value2onnx_parameter[node.outputs[0]].onnx_name], None)
-
+    elif isinstance(node.left, values.StrValue):
+        # Constant propagation, string concatenation, etc.
+        pass
     else:
         if node.binop == nodes.BinOpType.FloorDiv:
 
--- a/chainer_compiler/elichika/parser/functions_builtin.py
+++ b/chainer_compiler/elichika/parser/functions_builtin.py
@@ -56,7 +56,7 @@ class RangeFunction(functions.FunctionBase):
         if option._for_unroll:
             for ref in args.inputs:
                 if not ref.get_value().has_constant_value():
-                    assert False
+                    assert False, 'Loop unrolling was requested for non-constant sequence at %s' % line
 
             refs = []
             for num in range(*(ref.get_value().internal_value for ref in args.inputs)):
@@ -243,19 +243,23 @@ class ValuesFunction(functions.FunctionBase):
 
 
 class VEvalOptionFunction(functions.FunctionBase):
-    def __init__(self, func, flags = None):
+    def __init__(self, func):
         super().__init__()
         self.name = func.__name__
         self.args.analyze_args(func)
-        self.flags = flags
 
     def vcall(self, module: 'values.Field', graph: 'graphs.Graph', inst: 'values.ValueRef', args: 'functions.FunctionArgInput',
               option: 'vevaluator.VEvalOption' = None, line=-1):
         assert(inst is None)
 
         funcArgs = self.args.merge_inputs(inst, args)
-        if self.flags is not None:
-            self.flags.append(self.name)
+        args = []
+        for value in funcArgs.get_value().inputs:
+            assert value.has_constant_value(), "Arguments for elichika.flags were non-constant at %s" % line
+            args.append(value.internal_value)
+
+        if option is not None:
+            option.flags_cache.append((self.name, args))
 
         return values.ValueRef(values.NoneValue())
 
--- a/chainer_compiler/elichika/parser/functions_builtin.py
+++ b/chainer_compiler/elichika/parser/functions_builtin.py
@@ -242,3 +242,34 @@ class VEvalOptionFunction(functions.FunctionBase):
             self.flags.append(self.name)
 
         return values.ValueRef(values.NoneValue())
+
+
+class GetAttrFunction(functions.FunctionBase):
+    def __init__(self):
+        super().__init__()
+        self.name = 'getattr'
+        self.args.add_arg('object', None)
+        self.args.add_arg('name', None)
+
+    def vcall(self, module: 'Field', graph: 'Graph', inst: 'values.ValueRef', args: 'functions.FunctionArgInput', line=-1):
+        func_args = self.args.merge_inputs(inst, args)
+        name = func_args.get_value().get_value(key='name')
+        obj = func_args.keywords['object']
+
+        attr = obj.get_field().get_attribute(name.internal_value, graph.root_graph, False)
+
+        # property(getter)
+        if attr.has_obj() and isinstance(attr.get_ref().get_value(), values.FuncValue) and attr.get_ref().get_value().func.is_property:
+            func_value = attr.get_ref().get_value()
+            ret = func_value.func.vcall(func_value.module, graph, func_value.obj, functions.FunctionArgInput(), lineprop)
+            return ret
+
+        if attr.has_obj():
+            return attr
+
+        # if attr is not found
+        gotten_obj = obj.try_get_and_store_obj(name.internal_value, graph.root_graph)
+        if gotten_obj is not None:
+            return obj.get_field().get_attribute(name.internal_value, graph.root_graph, False)
+
+        return None
--- a/chainer_compiler/elichika/parser/functions_builtin.py
+++ b/chainer_compiler/elichika/parser/functions_builtin.py
@@ -178,7 +178,7 @@ class KeysFunction(functions.FunctionBase):
         if inst.in_container:
             raise Exception('Invalid operation')
 
-        keys = inst.get_value().internal_keys
+        keys = inst.get_value().internal_keys.values()
 
         vargs = []
         vargs_value = []
@@ -208,11 +208,11 @@ class ValuesFunction(functions.FunctionBase):
         if inst.in_container:
             raise Exception('Invalid operation')
 
-        key_refs = inst.get_value().internal_keys
+        key_hashes = inst.get_value().internal_keys.keys()
         attributes = inst.get_value().internal_values
         vargs = []
-        for key in key_refs:
-            varg = attributes.get_attribute(key.get_value().encode())
+        for hash in key_hashes:
+            varg = attributes.get_attribute(hash)
             if varg.has_obj():
                 vargs.append(varg.get_ref().get_value())
             else:
--- a/chainer_compiler/elichika/parser/functions_builtin.py
+++ b/chainer_compiler/elichika/parser/functions_builtin.py
@@ -11,22 +11,6 @@ import chainer.links as L
 def create_return_value_in_chainer_function():
     return values.TensorValue()
 
-def try_get_ref(value, name, lineprop) -> 'values.ValueRef':
-    if value is None:
-        utils.print_warning('Failed to get value in "{}".'.format(name), lineprop)
-        return None
-
-    if isinstance(value, values.Value):
-        assert(False)
-
-    if isinstance(value, values.Attribute):
-        if value.has_obj():
-            return value.get_ref()
-
-    if isinstance(value, values.ValueRef):
-        return value
-
-    return None
 
 class ChainerFunction(functions.FunctionBase):
     def __init__(self, func, ret_value_func = create_return_value_in_chainer_function):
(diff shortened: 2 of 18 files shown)


Rishav1:
PR is ready for review.

chainer_compiler/elichika/onnx_converters.py
```diff
@@ -317,7 +317,9 @@ def convert_node_bin_op(onnx_graph, node: 'nodes.NodeBinOp'):
         seq_right = right.create_sequence()
         onnx_graph.add_node(binops[node.binop], [seq_left, seq_right], [
                             value2onnx_parameter[node.outputs[0]].onnx_name], None)
-
+    elif isinstance(node.left, values.StrValue):
+        # Constant propagation, string concatenation, etc.
+        pass
```
shinh: Cannot we raise `NotImplementedError` here? Silently ignoring expressions would be confusing. If we can change `pass` to `raise` here, I think it'd be better to change the above comment to a TODO comment.

Or, this `pass` is actually safe since we'll raise exception later in veval_bin_op? If it's the case, could you update the above comment to explain why it's OK to ignore binops for strings?

chainer_compiler/elichika/parser/functions_builtin.py
```diff
@@ -47,22 +50,37 @@ def __init__(self):
...
-        value.name = '@F.{}.{}'.format(line, self.name)
-        node.set_outputs([value])
-        return values.ValueRef(value)
+    def vcall(self, module: 'values.Field', graph: 'graphs.Graph', inst: 'values.ValueRef', args: 'functions.FunctionArgInput',
+              option: 'vevaluator.VEvalOption' = None, line=-1):
+
+        if option._for_unroll:
+            for ref in args.inputs:
+                if not ref.get_value().has_constant_value():
+                    assert False
```
shinh: How about `assert False, 'Loop unrolling was requested for non-constant sequence at %s' % lineinfo`? (If it's not easy to obtain line info here, you can skip it for now)

chainer_compiler/elichika/parser/vevaluator.py
```diff
@@ -81,18 +72,44 @@ def c(self, value) -> 'AstContext':
...
     def __init__(self):
-        self.eval_as_written_target = False
+        self.history = []
+        self.flags = {
+            "eval_as_written_target": False,
+            "ignore_branch": False,
+            "for_unroll": False
+        }
+
+        deque(map(lambda flag: setattr(self.__class__, '_' + flag, property(fset=lambda obj, value: VEvalOption.generic_setter(obj, flag, value),
```
shinh: Why deque? I guess this is to evaluate the lazy sequence and `list` would be enough.

However, I'd recommend you to write this with a simple for-loop. Short loop comprehensions or map+lambda is sometimes improve readability (and it's actually fun to write :). IMHO however, it makes code harder to read or debug. For example, it'll be harder to insert print functions for debugging.

I like the guide of google: https://google.github.io/styleguide/pyguide.html#27-comprehensions--generator-expressions

chainer_compiler/elichika/parser/vevaluator.py
```diff
@@ -1173,105 +1213,185 @@ def veval_ast_for(astc : 'AstContext', local_field : 'values.Field', graph : 'Gr
...
+def veval_ast_continue(astc : 'AstContext', local_field : 'values.Field', graph : 'Graph', option : 'VEvalOption' = None):
     assert(isinstance(astc.nast, gast.gast.Continue))
     return None
 
-def veval_ast_break(astc : 'AstContext', local_field : 'values.Field', graph : 'Graph'):
+def veval_ast_break(astc : 'AstContext', local_field : 'values.Field', graph : 'Graph', option : 'VEvalOption' = None):
     assert(isinstance(astc.nast, gast.gast.Break))
     return None
 
+def veval_ast_with(astc : 'AstContext', local_field : 'values.Field', graph : 'Graph', option : 'VEvalOption' = None):
```
shinh: (TODO for myself: understand this)

chainer_compiler/elichika/onnx_converters.py
Rishav1: veval_bin_op executes before onnx_converters. Since binary operations on strings types are being done in veval_bin_op itself (and incompatible types are raising assertion errors there), the generated NodeBinOp node doesn't have to do anything and hence is no-op. No need to raise any `NotImplementedError`.

chainer_compiler/elichika/parser/functions_builtin.py
Rishav1: Yeah, I guess that is a good idea.

chainer_compiler/elichika/parser/vevaluator.py
Rishav1: Yes, `list` would be enough I suppose. I was initially using a for loop only, but it had some bug that I couldn't pin point. These two code snippets had different behaviors. The first one worked as expected, but the second one wasn't setting the correct flag on calling `option.<flag>()` (It always was setting `option._for_unroll`  to `True`, probably because it is the last `flag` in the for loop). 

```python
list(map(lambda flag: setattr(self.__class__, '_' + flag, property(fset=lambda obj, value: VEvalOption.generic_setter(obj, flag, value),
                                                                    fget=lambda obj: VEvalOption.generic_getter(obj, flag))),
            self.flags.keys()))
list(map(lambda flag: setattr(self.__class__, flag, auto_set_unset(lambda obj: VEvalOption.generic_setter(obj, flag, True), flag)),
            self.flags.keys()))
```

```python
for flag in self.flags.keys():
    setattr(self.__class__, '_' + flag, property(fset=lambda obj, value: VEvalOption.generic_setter(obj, flag, value),
                                                    fget=lambda obj: VEvalOption.generic_getter(obj, flag)))

    setattr(self.__class__, flag, auto_set_unset(lambda obj: VEvalOption.generic_setter(obj, flag, True), flag))
```

Rishav1: I suspected that this behavior was due to the fact that lambda expression behaves differently than what is expected.
```python
>>> x = []
>>> for i in range(10):
...     x.append(lambda: i)
... 
>>> x[0]()
9
```

chainer_compiler/elichika/parser/vevaluator.py
Rishav1: A description of how I designed the support for `With` syntax with flags.
- veval_ast(astc<gast.gast.withitem>, ...) behaves almost like `veval_ast_assign` assigning the output of processed `astc.nast.context_expr` to attribute `astc.nast.optional_vars`. Before assigning the output of `astc.nast.context_expr` however, we process  `__enter__` function of the context_expr attribute and use the output in place of it. This handles the case where the custom __enter__ function doesn't return self object.

- Compiler flags such as `with flags.for_unroll()` is handled by adding their function names to `builtin_function_converters`, with endpoint `VEvalOptionFunction`. This class extends `FunctionBase` and it's `vcall` just sets the corresponding flag of VEvalOption object option (through a shared `vevaluator.common_flags_cache`). (Note: I can probably remove this cache and modify the option object directly since vcall has access to it after the interface change I did later.)

-  veval_ast(astc<gast.gast.with>, ...) processes the body after all the withitems have been processed and then calls the withitems' context_expr' `__exit__` functions. The purpose of `ExitStack` is to allow multiple `with option.<flag>()` succintly (L1241). The option flags are set only in the with scope(In this case `ExitStack` scope) as they are unset in `__exit__` call of VEvalOption, thanks to `auto_set_unset`.

shinh: Thanks for you explanation!

> I can probably remove this cache and modify the option object directly since vcall has access to it after the interface change I did later.

This sounds great

testcases/elichika_tests/syntax/With.py
```diff
@@ -0,0 +1,110 @@
...
+        super(ForUnroll, self).__init__()
+        self.x1 = 1
+        self.x2 = 2
+        self.x3 = 3
+        self.dict = {'x1': 1, 'x2': 2, 'x3': 3}
+
+    def forward(self):
+        ret = 0
+
+        with flags.for_unroll():
```
shinh: Can we extend `for_unroll` to take an optional boolean argument which makes it possible for users to go back to normal mode? A usecase scenario in my mind:

```
with flags.for_unroll():
  hidden = inputs  # A list with `seq_length` elements.
  for layer_id in range(num_layers):  # num_layers is a constant
    next_hidden = []
    with flags.for_unroll(unroll=False):
      state = rnn_initial_state()
      for time in range(seq_length):  # seq_length is not a constant
         next_h, state = self['rnn_%d' % layer_id](hidden[time], state)
         next_hidden.append(next_h)
    hidden = next_hidden
```

chainer_compiler/elichika/onnx_converters.py
Rishav1: Yes, it should pass.

chainer_compiler/elichika/parser/vevaluator.py
Rishav1: Okay, I will use `list`. `functools.partial` seems like a nice solution, but it changes the argument list of the lambda function, which would be unwanted in this case as the lambda function is being used as setter/getter functions which require a fixed argument list (obj, key, [value]).

chainer_compiler/elichika/onnx_converters.py
Rishav1: It is passing already without emitting any onnx node, actually. Because elichika does the constant propagation while parsing.

shinh: Hmm? I got the following error with this PR and my modification above
```
$ ./build/tools/run_onnx out/elichika_syntax_Print
Initializing ChainerX...
Loading model...
Loading data...
Found 1 test cases
Constructing model...
Failed to schedule (): ChainerPrint(@C_test_1, _L.8) -> ()
 _L.8 cannot be ready
Failed graph is stored in /tmp/chainer_compiler_failure.onnx
Check `false' failed! in CheckSanity at ../compiler/scheduler.cc:233:
zsh: abort (core dumped)  ./build/tools/run_onnx out/elichika_syntax_Print
```

Rishav1: That is strange. For me, the test is passing just fine.

```
test 10
foo test 3
*** Testing out/elichika_tmp/tmp ***
build/tools/run_onnx --test out/elichika_tmp/tmp
Initializing ChainerX...
Loading model...
Loading data...
Found 1 test cases
Constructing model...
Generate code...
Running for out/elichika_tmp/tmp/test_data_set_0
Verifying the result...
OK: #returned_value
Elapsed: 0.285 msec
OK!
```

testcases/elichika_tests/syntax/With.py
Rishav1: I added `unroll=False`. The above example however won't work without inside loop being unrolled as `'rnn_%d' % layer_id` requires constant propagation which would only work with unrolled for loop currently.

chainer_compiler/elichika/onnx_converters.py
shinh: This also failed on CI? https://github.com/pfnet-research/chainer-compiler/pull/481/commits/6ba836bdbde824309a12aa0f723a4deece082424

testcases/elichika_tests/syntax/With.py
shinh: Is this easy to do constant propagation in outside `for_unroll`?

chainer_compiler/elichika/onnx_converters.py
shinh: Sorry, it had a different change. Updated link: https://github.com/pfnet-research/chainer-compiler/pull/481/commits/2ba3dc7810380d25ec461182885ef343551d8582

https://travis-ci.org/pfnet-research/chainer-compiler/builds/557706471?utm_source=github_status&utm_medium=notification

Rishav1: I found the bug. Will submit a PR for the fix.

PR #437, opened 2019-07-01
Modify examples so that arbitrary batch-size can be specified when training

ir5:
`python examples/mnist/train_mnist.py --compile mymodel.onnx -b 101` will become possible.

### DIFF
--- a/examples/imagenet/train_imagenet_multi.py
+++ b/examples/imagenet/train_imagenet_multi.py
@@ -68,6 +68,22 @@ class ValTransform(object):
         return img, label
 
 
+class FixedBatchDataset(chainer.dataset.DatasetMixin):
+    # Make the dataset size multiple of the batch-size by augmentation
+
+    def __init__(self, dataset, batchsize):
+        self.dataset = dataset
+        self.batchsize = batchsize
+        d = len(self.dataset)
+        self._len = ((d + batchsize - 1) // batchsize) * batchsize
+
+    def __len__(self):
+        return self._len
+
+    def get_example(self, idx):
+        return self.dataset[idx % len(self.dataset)]
+
+
 def main():
     model_cfgs = {
         'resnet50': {'class': ResNet50, 'score_layer_name': 'fc6',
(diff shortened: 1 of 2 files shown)


examples/imagenet/train_imagenet_multi.py
```diff
@@ -68,6 +68,22 @@ def __call__(self, in_data):
...
+        self.dataset = dataset
+        self.batchsize = batchsize
+        d = len(self.dataset)
+        self._len = ((d + batchsize - 1) // batchsize) * batchsize
+
+    def __len__(self):
+        return self._len
+
+    def get_example(self, idx):
+        return self.dataset[idx % len(self.dataset)]
```
shinh: I'm a bit worried about the bias from this. If I understand correctly, this oversample first data twice as many times as other data? Not sure if this is a great idea, but how about returning a data in random index when idx overflows?

examples/imagenet/train_imagenet_multi.py
```diff
@@ -181,6 +197,8 @@ def main():
     val_indices = chainermn.scatter_dataset(val_indices, comm, shuffle=True)
     train_data = train_data.slice[train_indices]
     val_data = val_data.slice[val_indices]
+    train_data = FixedBatchDataset(train_data, args.batchsize)
```
shinh: Could you introduce a flag which enable FixedBatchDataset?

examples/imagenet/train_imagenet_multi.py
ir5: I noticed that we can specify an "ignore label" for the over-sampled data for the classification problem. This will cause (almost?) no bias in both training and validation.

examples/imagenet/train_imagenet_multi.py
ir5: Implemented.

ir5: I put this flag to be true by default, because I don't want to stop the training process by error after epoch 1 finished.

examples/imagenet/train_imagenet_multi.py
```diff
@@ -68,6 +68,27 @@ def __call__(self, in_data):
...
+
+    def __len__(self):
+        return self._len
+
+    def get_example(self, idx):
+        if idx < len(self.dataset):
+            return self.dataset[idx]
+        else:
+            x_dummy, _ = self.dataset[0]
+            t_dummy = -1
```
shinh: Ah, nice! Could you add a link to https://docs.chainer.org/en/stable/reference/generated/chainer.functions.softmax_cross_entropy.html to explain what -1 is? Also, it'd be better add `self.ignore_label`, e.g.,:

```
# `ignore_label` should be consistent with
# https://docs.chainer.org/en/stable/reference/generated/chainer.functions.softmax_cross_entropy.html
def __init__(self, dataset, batchsize, ignore_label=-1):
  self.ignore_label = ignore_label
```

ir5: Thank you for comment. Modified the code.

PR #440, opened 2019-07-02
Add documentation for training from python

ir5:


shinh:
Great, thanks! Yeah, I agree the usage is complicated and we really need a doc

PR #424, opened 2019-06-28
Add script to quantize ONNX model

take-cheeze:
https://github.com/microsoft/onnxruntime/tree/master/onnxruntime/python/tools/quantization

PR #442, opened 2019-07-03
Add shape evaluators for quantized ops

take-cheeze:
rel: #424
Seems some shape of nodes created by [quantize.py](https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/python/tools/quantization/quantize.py) is incomplete.


shinh:
Do you know why ONNX's standard shape inference does not work for these ops?

take-cheeze:
Seems this was caused by these:
https://github.com/pfnet-research/chainer-compiler/blob/23674585ac45159fcd4c93418015af6f14be8780/compiler/graph_builder.cc#L52
https://github.com/pfnet-research/chainer-compiler/blob/f7a3a226e4fbd09767204bcd74e0b6c5e2c420ee/compiler/graph.cc#L317
Will move this configurable from chxvm.json.

shinh:
:( Sorry for these hard-coded numbers..

PR #452, opened 2019-07-05
Do not depend on CMakeLists.txt from tgz packages

shinh:


shinh:
I confirmed changing commandline re-triggers wget even with make backend. Also, it seemed wget always runs without this change.

PR #453, opened 2019-07-05
refactor: commenting out noop lines in ESP Net.

Rishav1:
Passes ESP Net ONNX generation.


shinh:
I think this is fine. This TODO can be removed when #465 is fixed.

PR #455, opened 2019-07-05
Free BN output

ir5:


shinh:
ChainerCI test this please

Just in case...

PR #460, opened 2019-07-05
Plumb ChxVMOption::dump_outputs_dir from Python

shinh:


shinh:
ChainerCI test this please

PR #466, opened 2019-07-06
Count memory consumption of ChxVM inputs/outputs

shinh:


shinh:
ChainerCI, test this please

PR #477, opened 2019-07-09
Do not use dldt for Resize on CPU

shinh:


shinh:
ChainerCI, test this please

PR #478, opened 2019-07-09
Fix graph construction of computation_order

ir5:
I used a little dirty workaround such as `chainer_order_offset` and `retained.insert({new_value, new_value});`. I'm not sure if these can be written in more sophisticated way...


shinh:
ChainerCI test this please

PR #484, opened 2019-07-12
Bump up chainer version

take-cheeze:
to 23941e89ea4725f63a53ac532d51b521352dd506

PR #485, opened 2019-07-12
Use ChainerX softmax and bit shifts

take-cheeze:
rel: #484


take-cheeze:
ChainerCI, test this please!

PR #494, opened 2019-07-17
Add options to ease checks

take-cheeze:


take-cheeze:
ChainerCI, test this please!

tools/run_onnx.cc
```diff
@@ -414,7 +414,7 @@ class ModelRunner {
     std::vector<std::string> backprop_ins_;
 };
 
-void VerifyOutputs(const InOuts& outputs, const TestCase& test_case, const cmdline::parser& args, bool strict_check) {
+void VerifyOutputs(const InOuts& outputs, const TestCase& test_case, const cmdline::parser& args, bool strict_check, bool show_diff) {
```
shinh: My bad, I think `strict_check` wasn't a good name. How about renaming this to `check_values`?

tools/run_onnx.cc
```diff
@@ -534,6 +537,8 @@ void RunMain(const std::vector<std::string>& argv) {
     args.add<double>("atol", '\0', "atol of AllClose", false, 1e-6);
     args.add("equal_nan", '\0', "Treats NaN equal");
     args.add("no_catch", '\0', "Do not catch the exception in ChxVM for better GDB experience");
+    args.add("no_strict_check", '\0', "Disable strict check of node output");
```
shinh: `no_check_values`?

tools/run_onnx.cc
```diff
@@ -534,6 +537,8 @@ void RunMain(const std::vector<std::string>& argv) {
...
+    args.add("show_diff", '\0', "Show diff of strict check without stopping");
```
shinh: How about `always_show_diff`? Also, how about "Show diff even when value check is skipped" for the help message?

PR #496, opened 2019-07-22
Implementation of "graph theoretic framework of recomputation algorithm"

ir5:
I implemented ["A Graph Theoretic Framework of Recomputation Algorithms"](https://arxiv.org/abs/1905.11722). 

* For the estimation of available memory budget, I used `cudaMemGetInfo`. This adds a dependency to CUDA in `compiler` directory.
* Though I checked that the memory reduction works in a simple case, this implementation may contain some bug.


compiler/computation_order/policy_gt.cc
```diff
@@ -0,0 +1,387 @@
...
+int64_t AutomaticBudgetDetection() {
+#ifdef CHAINER_COMPILER_ENABLE_CUDA
+    size_t free, total;
+    if (cudaMemGetInfo(&free, &total) != cudaSuccess) {
+        return 100000000000LL;
+    }
+    const size_t used = total - free;
+    return total - 3 * used;
+#else
+    return nonstd::nullopt;
```
take-cheeze:
```suggestion
    return absl::nullopt;
```
We are using absl like chainerx now.

compiler/computation_order/policy_gt.cc
```diff
@@ -0,0 +1,387 @@
...
+            Node* producer = (*it)->producer();
+            CHECK(producer);
+            orders.emplace_back(Order::kComputeBackward, producer, nullptr);
+        }
+    }
+
+    return orders;
+}
+
+int64_t AutomaticBudgetDetection() {
```
shinh: Maybe it's enough to just include `runtime/meminfo.h` and use `GetMemoryUsageInBytes` here?

compiler/CMakeLists.txt
```diff
@@ -56,6 +56,7 @@ include_directories(${GSLLITE_INCLUDE_DIRS})
 include_directories(${CHAINER_COMPILER_ROOT_DIR})
 include_directories(${CMAKE_CURRENT_BINARY_DIR}/..)
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
+include_directories(${CUDA_INCLUDE_DIRS})
```
shinh: Unnecessary if we use runtime/meminfo instead?

compiler/computation_order/policy_gt.cc
ir5: Thank you, fixed.

compiler/CMakeLists.txt
ir5: When I tried to use runtime/memnifo, I encountered an issue that `g_meminfo_enabled` may be `false` and I cannot call `GetMemoryUsageInBytes()` properly. If it is ok to set `g_meminfo_enabled=true` before calling `GetMemoryUsageInBytes()`, then I think I can change this part.

compiler/computation_order/policy_gt.cc
```diff
@@ -0,0 +1,380 @@
...
+    runtime::g_meminfo_enabled = true;
```
shinh: Ah, sorry for this hacky global variable. I'll remove this in future.

PR #506, opened 2019-07-23
Set ignore_gitignore to fetch ONNX tests

shinh:


shinh:
ChainerCI, test this please

PR #518, opened 2019-07-25
Fix skip_inference option in python

ir5:
#508 caused a bug: computation_order option will not work from python code.
This PR will fix the issue.

### DIFF
--- a/chainer_compiler/chainer_compiler.py
+++ b/chainer_compiler/chainer_compiler.py
@@ -311,7 +311,7 @@ class CompiledModel(chainer.Chain):
 
         runtime_kwargs = {}
         if (self.runtime_kwargs is not None and
-            self.num_iterations % (self.quiet_period + 1) == 0):
+           self.num_iterations % (self.quiet_period + 1) == 0):
             runtime_kwargs.update(self.runtime_kwargs)
         self.num_iterations += 1
 
--- a/chainer_compiler/chainer_compiler.py
+++ b/chainer_compiler/chainer_compiler.py
@@ -311,7 +311,7 @@ class CompiledModel(chainer.Chain):
 
         runtime_kwargs = {}
         if (self.runtime_kwargs is not None and
-           self.num_iterations % (self.quiet_period + 1) == 0):
+            self.num_iterations % (self.quiet_period + 1) == 0):
             runtime_kwargs.update(self.runtime_kwargs)
         self.num_iterations += 1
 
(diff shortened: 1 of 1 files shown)


chainer_compiler/chainer_compiler.py
```diff
@@ -238,10 +238,8 @@ def __init__(self, model, onnx_file, used_translator, dump_onnx=False,
 
     def compile(self, onnx_file):
         # TODO(hamaji): Revive shape inference.
```
shinh: Could you move this comment to line 262 again?

chainer_compiler/chainer_compiler.py
```diff
@@ -262,6 +260,8 @@ def compile(self, onnx_file):
                              bwd_graph.dump() +
                              '\n=== ^^^ backward ^^^ ===\n')
 
+        _chainer_compiler_core.configure(skip_inference=True)
```
shinh: I think the original logic,
```
        compiler_kwargs = {'skip_inference': True}
        if self.compiler_kwargs is not None:
            compiler_kwargs.update(self.compiler_kwargs)
        _chainer_compiler_core.configure(**compiler_kwargs)
```
is more appropriate here. Doesn't this work for you?

chainer_compiler/chainer_compiler.py
```diff
@@ -311,7 +311,7 @@ def forward(self, *args):
 
         runtime_kwargs = {}
         if (self.runtime_kwargs is not None and
-            self.num_iterations % (self.quiet_period + 1) == 0):
+           self.num_iterations % (self.quiet_period + 1) == 0):
```
shinh: Could you revert this unnecessary change?

chainer_compiler/chainer_compiler.py
ir5: Because `graph.backward_to_with_order` depends on some global variables, `_chainer_compiler_core.configure` should be called before `graph.backward_to_with_order`.

chainer_compiler/chainer_compiler.py
ir5: fixed.

chainer_compiler/chainer_compiler.py
ir5: It seems that `flake8` outputs an error at this part. But if we don't need to care, then I will revert.

chainer_compiler/chainer_compiler.py
shinh: Yeah, I'm OK with line 242, but I think it's better to put the 4 lines here again. Otherwise, `compile` calls in line 270 and 271 will ignore `compiler_kwargs` specified by the user.

ir5: Ah, I understood. Thank you for pointing out this part. Now I fixed.

shinh:
Thanks!!

PR #519, opened 2019-07-25
Fix --check_(infs|nans) for optional outputs

shinh:


shinh:
ChainerCI, test this please

PR #529, opened 2019-07-27
Compile resnet

Rishav1:


Rishav1:
Found the cause of travis failure. Implemented builtin `sum` in this PR and elichika is confusing `F.sum` with the builtin `sum`.

durswd:
it maybe fix this problem.

https://github.com/pfnet-research/chainer-compiler/pull/530

Rishav1:
Some issues in support for max, min & sum builtin. Will add it in a different PR because they are unrelated to resnet.

PR #533, opened 2019-07-28
Rename all XC prefixes to ChxVM

shinh:


shinh:
ChainerCI, test this please

PR #536, opened 2019-07-28
Fix to avoid to install onnxruntime

disktnk:
`third_party/onnx-chainer[test]`: install pytest and ChaienrCV
`third_party/chainer` -> `third_party/chainer[test]`: to install pytest supported by Chainer before installing ONNX-Chainer


disktnk:
ChainerCI, test this please

PR #544, opened 2019-07-31
[WIP] Support Windows

durswd:


compiler/computation_order/policy_gt.cc
```diff
@@ -229,7 +229,8 @@ std::tuple<int64_t, int64_t, size_t> ComputeConsumptionInfo(const SimpleGraph& s
 std::vector<NodeSet> ComputeDP(const SimpleGraph& sg, const std::vector<NodeSet>& lower_sets, const int64_t& budget) {
     size_t nl = lower_sets.size();
     // opt[lower_set_index][flops] := <minimum memory consumption, prev_ls, prev_flops>
-    std::map<size_t, std::tuple<int64_t, size_t, size_t>> opt[nl];
+    std::vector<std::map<size_t, std::tuple<int64_t, size_t, size_t>>> opt;
```
shinh: `opt(nl)` and remove the next line?

runtime/CMakeLists.txt
```diff
@@ -3,13 +3,17 @@ set(CHAINER_COMPILER_COMMON_DIR ${CHAINER_COMPILER_ROOT_DIR}/common)
 set(GOOGLETEST_INCLUDE_DIRS ${CHAINER_COMPILER_ROOT_DIR}/googletest/googletest/include)
 set(GSLLITE_INCLUDE_DIRS ${CHAINER_COMPILER_ROOT_DIR}/gsl-lite/include)
 
+if(MSVC)
+    find_package(PythonInterp 3)
```
shinh: Thanks for pointing this out! I think we should move this to the top-level CMakeLists.txt without `if(MSVC)` guard and replace all `python3`.

PR #546, opened 2019-07-31
Start supporting SNPE backend

take-cheeze:
- [x] Run dlc like caffe2 https://github.com/pytorch/pytorch/tree/b8c8977be797cb16354aa97c3fcdc910d3e9e5c5/caffe2/mobile/contrib/snpe
- [x] Generate ONNX to dlc


runtime/ops/dldt.cc
```diff
@@ -9,14 +9,11 @@
 
 #include <inference_engine.hpp>
 
-#include <common/log.h>
-
 #else
```
shinh: could you remove #else?

PR #552, opened 2019-08-01
Fix tiny yolo v2 path

take-cheeze:


take-cheeze:
Seems to have Tiny YOLO v2 model generated by ONNX in https://github.com/pfnet-research/chainer-compiler/blob/35afb6f33c5a84630e89bd00eb2b4c6ee47e8a94/scripts/gen_chainercv_model_tests.py#L46 already

take-cheeze:
Will split some needs to be merge so closing

PR #558, opened 2019-08-02
Fix type issue in ROI

### LINKED ISSUE #545: RoiAlign batch_indices type handled incorrectly
Currently we expect int32_t tensor: https://github.com/pfnet-research/chainer-compiler/blob/38973e93910f9d15e4b225116944cf088e7bb23c/runtime/ops/roi.cc#L192
Though ONNX spec expects to be int64_t tensor: https://github.com/onnx/onnx/blob/98f72629bde7285f5a601f95d122ce43cea97a9c/docs/Operators.md#inputs-89

take-cheeze:
closes #545

### DIFF
--- a/runtime/ops/roi.cc
+++ b/runtime/ops/roi.cc
@@ -96,16 +96,16 @@ absl::optional<std::tuple<double, int64_t, int64_t>> get_bounds(double p, int64_
 using ArrayIndices = chainerx::StackVector<int64_t, chainerx::kMaxNdim>;
 template <typename T>
 T& ContiguousArrayAt(chainerx::Array& a, const ArrayIndices& indices) {
-    assert(a.IsContiguous());
-    assert(a.shape().size() == indices.size());
-    assert(a.dtype() == chainerx::PrimitiveType<T>::kDtype);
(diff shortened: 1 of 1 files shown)


runtime/ops/roi.cc
```diff
@@ -189,7 +189,12 @@ class ROIAlign2DImpl {
         std::vector<PixelPos> pixel_x(pooled_width * roi_bin_grid_w);
         std::vector<PixelPos> pixel_y(pooled_height * roi_bin_grid_h);
 
-        int64_t roi_batch_ind = ContiguousArrayAt<int32_t>(contiguous_bottom_roi_indices, {n});
+        int64_t roi_batch_ind;
+        if (contiguous_bottom_roi_indices.dtype() == chainerx::Dtype::kInt64) {
+            roi_batch_ind = ContiguousArrayAt<int64_t>(contiguous_bottom_roi_indices, {n});
+        } else {
```
shinh:
```
else if (... Int32) {
 ...
} else {
  CHECK(false) << ...dtype();
}
```
please?

PR #561, opened 2019-08-03
Support Windows

durswd:


tools/dump.cc
```diff
@@ -69,12 +80,24 @@ void RunMain(int argc, char** argv) {
             // ONNX test directory.
             DumpONNX(filename + "/model.onnx", args);
 
+            std::vector<std::string> filenames;
+#if _WIN32
+            fs::directory_iterator iter(filename);
+
+            std::regex filter(filename + "/*/*.pb");
```
shinh: I don't think this regexp is equivalent to the original code. Maybe something like `"/[^/]+/[^/]+\\.pb"` ?

CMakeLists.txt
```diff
@@ -10,6 +10,18 @@ list(INSERT CMAKE_MODULE_PATH 0 ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
 include(third-party)
 
+find_package(PythonInterp 3)
```
shinh: This a dup

common/codegen_util.py
```diff
@@ -28,3 +28,18 @@ def cond(conds, bodies):
     lines += bodies[len(conds)]
     lines.append('}')
     return lines
+
+def cond_msvc(flag, conds, bodies):
```
shinh: Cannot we unconditionally use this?

compiler/nvrtc_builder.cc
```diff
@@ -21,9 +22,10 @@ namespace chainer_compiler {
 namespace {
 
 std::string CleanseIdent(const std::string& s, const char* prefix = "v_") {
+    std::locale loc;
     std::string o = prefix;
     for (char c : s) {
-        if (std::isalnum(c)) {
+        if (std::isalnum(c, loc)) {
```
shinh: Not sure why we need this change: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/isalnum-iswalnum-isalnum-l-iswalnum-l?view=vs-2019

compiler/util.cc
```diff
@@ -55,8 +55,9 @@ void StripONNXModel(onnx::ModelProto* model) {
 
 std::string CleanseIdent(const std::string& s) {
     std::string o;
+    std::locale loc;
     for (char c : s) {
-        if (std::isalnum(c)) {
+        if (std::isalnum(c, loc)) {
```
shinh: ditto

PR #565, opened 2019-08-06
fixed for windows

durswd:
rename variables

### DIFF
--- a/menoh/CMakeLists.txt
+++ b/menoh/CMakeLists.txt
@@ -29,7 +29,6 @@ target_link_libraries(menoh_chainer_compiler
   onnx
   onnx_proto
   ${PROTOBUF_LIBRARY}
-  pthread
   ${CHAINER_COMPILER_NGRAPH_LIBRARIES}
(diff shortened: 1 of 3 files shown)


scripts/generate_flags_code.py
```diff
@@ -179,6 +179,14 @@
 
 elif args.mode == 'compiler_flags.cc':
     f.write('''
+
+#ifdef _WIN32
+// HACK for Windows including order
+#define NOMINMAX
```
shinh: Could you check if `WIN32_LEAN_AND_MEAN` works?

menoh/CMakeLists.txt
```diff
@@ -38,6 +37,12 @@ target_link_libraries(menoh_chainer_compiler
   absl::optional
   )
 
+if (!WIN32)
+  target_link_libraries(menoh_chainer_compiler
+  pthread
```
shinh: I think we should define `${CHAINER_COMPILER_PTHREAD_LIBRARIES}` or something in the toplevel CMakeFiles.txt

shinh:
I'll try doing these follow ups

PR #575, opened 2019-08-08
Win32 followup 2nd round

shinh:
- Always use cond generator for win32
- Minor style fixes


shinh:
ChainerCI, test this please

PR #576, opened 2019-08-08
fixed isalnum for windows

durswd:


compiler/nvrtc_builder.cc
```diff
@@ -20,10 +20,9 @@ namespace chainer_compiler {
 namespace {
 
 std::string CleanseIdent(const std::string& s, const char* prefix = "v_") {
-    std::locale loc;
     std::string o = prefix;
     for (char c : s) {
-        if (std::isalnum(c, loc)) {
+        if (std::isalnum(c)) {
```
shinh: Could you use the two-argument version? I decided to use the default `locale`, which is classic locale instead of system's.

compiler/util.cc
```diff
@@ -56,9 +55,8 @@ void StripONNXModel(onnx::ModelProto* model) {
 
 std::string CleanseIdent(const std::string& s) {
     std::string o;
-    std::locale loc;
     for (char c : s) {
-        if (std::isalnum(c, loc)) {
+        if (std::isalnum(c)) {
```
shinh: ditto

compiler/util.cc
```diff
@@ -1,9 +1,8 @@
 #include "compiler/util.h"
 
-#include <locale>
-
 #include <common/strutil.h>
 #include <compiler/tensor.h>
+#include <cctype>
```
shinh: Please move this (and locale) before `common/strutil.h`

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

PR #592, opened 2019-08-19
Add run onnx menoh

okdshin:


menoh/run_onnx.cpp
```diff
@@ -0,0 +1,458 @@
...
+bool g_quiet;
+
+#define LOG() \
+    if (!g_quiet) std::cerr
+
+namespace chainer_compiler {
+namespace runtime {
+namespace {
+
+bool IsDir(const std::string& filename) {
```
shinh: Please do not copy these utility functions?

menoh/run_onnx.cpp
```diff
@@ -0,0 +1,458 @@
...
+    struct stat st;
+    CHECK_EQ(0, stat(filename.c_str(), &st)) << "failed to stat: " << filename << ": " << strerror(errno);
+    return S_IFDIR == (st.st_mode & S_IFMT);
+}
+
+chainerx::Array MakeArrayFromONNX(const onnx::TensorProto& xtensor) {
```
shinh: I think it's probably better to add run_onnx_util.h or something to share the implementation of them.

okdshin: I see. I will modify it to share them so.

PR #602, opened 2019-08-21
Use memcpy instead of std::copy

shinh:
Since std::copy is usually much slower than memcpy.


okdshin:
Is it real? IMO, in C++, we must prefer C++ functions but not C functions.
Do you have any evidence which shows replacing std::copy with memcpy is really effective and is not a premature optimization?

shinh:
I was pretty sure this is the right change since I've ever done similar replacement several times. However, today I learned libstdc++ can call `__builtin_memmove` (which is only slightly slower than memcpy, I think). I should update my knowledge based on old experience :(

PR #603, opened 2019-08-21
Add check for menoh api

okdshin:


menoh/menoh_chainer_compiler.cpp
```diff
@@ -749,6 +759,9 @@ menoh_error_code menoh_model_run(menoh_model_handle model) {
                 auto const& array = chainerx::AsContiguous(output.second->GetArray());
                 auto const& shape = array.shape();
                 auto bytesize = shape.GetTotalSize() * chainerx::GetItemSize(array.dtype());
+                assert(model->variable_profiles.find(output.first) != model->variable_profiles.end());
```
shinh: CHECK_NE please

menoh/menoh_chainer_compiler.cpp
```diff
@@ -749,6 +759,9 @@ menoh_error_code menoh_model_run(menoh_model_handle model) {
...
+                assert(bytesize == menoh_impl::total_size_in_bytes(variable_profiles.find(output.first)->second) &&
```
shinh: CHECK_EQ(...) << "allocated output..."

menoh/menoh_chainer_compiler.cpp
okdshin: OK. I will modify.

okdshin: I can't use CHECK_NE here because iterator is not printable.
I used `CHECK(a != b)` instead of `CHECK_NE(a, b)`

PR #604, opened 2019-08-21
Fix menoh filename

okdshin:
There is inconsistency about filenames in menoh directory i.e. there are `.cc` and `.cpp`.
This PR change `.cpp` to `.cc`


shinh:
Hmm if we do this, shouldn't we rename hpp to h for consistency and menoh.hpp will conflict with menoh.h?

menoh/menoh_chainer_compiler.cc
```diff
@@ -18,12 +18,12 @@
 #include <compiler/onnx.h>
 #include <compiler/passes.h>
 #include <compiler/util.h>
-#include <menoh/menoh_chainer_compiler_util.hpp>
 #include <runtime/chainerx_util.h>
 #include <runtime/chxvm.h>
 #include <runtime/chxvm.pb.h>
 #include <runtime/chxvm_var.h>
 #include <tools/util.h>
+#include <menoh/menoh_chainer_compiler_util.hpp>
```
shinh: I'm trying to follow ChainerX's coding style which is based on Google's (https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes). If you follow the rule, I think this include should be at the top of this .cc file

menoh/run_onnx.cc
```diff
@@ -51,6 +49,8 @@
 #include <tools/log.h>
 #include <tools/run_onnx_util.h>
 #include <tools/util.h>
+#include <menoh/menoh.hpp>
```
shinh: As for them, I'd revert this change or add a blank line between line 51 and 52.

menoh/menoh_chainer_compiler.cc
okdshin: It is my understanding that `menoh_chainer_compiler.cc` doesn't contain implementations for functions declared in `menoh_chainer_compiler_util.h` and also it is not test codes for them, so `menoh_chainer_compiler_util.h` is classified as `7. Your project's .h files`. Thus, the include line must be here. Is it incorrect?

shinh: Sorry, I missed "_util". I think the original position is consistent with other code since I'm treating all header files in this repository as "Your project's .h files". However, if you like, moving this include is also fine to me. Please insert a blank line between line 25 and 26 then.

okdshin: OK． I will add blank lines．

okdshin:
IMO, the name `menoh.hpp` itself is a part of interface of menoh. So it is OK, it breaks consistency with other filenames tho.
If you allow that inconsitency, I will keep the filename `menoh.hpp`  and add some comments about it.

okdshin:
We decided not to change the names.

PR #605, opened 2019-08-22
Bump up chainer version

ir5:
to 4381df84878e08f55f0cba31759fc7628120c449

This will enable to use hook functions in ChainerX memory pool.


scripts/large_models/chainer_chain.py
```diff
@@ -10,7 +10,7 @@
 from chainer.links.connection.convolution_2d import Convolution2D
 from chainer.links.connection.linear import Linear
 from chainer.functions.noise.dropout import dropout
-from chainer.functions.pooling.max_pooling_2d import max_pooling_2d
+from chainer.functions.pooling.max_pooling_nd import max_pooling_2d
```
shinh: This is fine, but cannot we use `chainer.functions import max_pooling_2d` so we can support both old chainer and new one?

ir5: Thank you, fixed.

Issue #612, opened 2019-08-23
Run Mask R-CNN R-50-FPN in model zoo


take-cheeze:
https://github.com/onnx/models/tree/master/vision/object_detection_segmentation/mask-rcnn

Blockers:
- ceil_mode for `MaxPool`
- `NonZero` op

shinh:
and NMS plumbing

shinh:
+TopK

take-cheeze:
👀  https://github.com/pfnet-research/chainer-compiler/pull/618

take-cheeze:
Scatter missing too: https://github.com/onnx/onnx/blob/master/docs/Operators.md#Scatter

take-cheeze:
Current state after https://github.com/pfnet-research/chainer-compiler/pull/639:
```
Initializing ChainerX...
Loading model...
Loading data...
Found 1 test cases
Constructing model...
Generate code...
Running for ~/dev/archive/mask_rcnn_R_50_FPN_1x/test_data_set_0
run_onnx: ~/dev/chainer-compiler/third_party/chainer/chainerx_cc/chainerx/cuda/cuda_device/indexing.cu:37: chainerx::Axes chainerx::cuda::_GLOBAL__N__43_tmpxft_000006b2_00000000_6_indexing_cpp1_ii_48d72df5::MakeRollingPermutation(int8_t, int8_t, int8_t): Assertion `first_axis < last_axis' failed.
/bin/bash: line 1: 11092 Aborted                 (core dumped) ./build/tools/run_onnx -d cuda:0 --test ~/dev/archive/mask_rcnn_R_50_FPN_1x
```

take-cheeze:
```
Unknown Tensor shape: 6803 producer=Expand
Unknown Tensor shape: 6831 producer=Expand
Unknown Tensor shape: 6748 producer=Cast
Unknown Tensor shape: 6776 producer=Cast
Unknown Tensor shape: 6804 producer=Cast
Unknown Tensor shape: 6832 producer=Cast
Unknown Tensor shape: 6749 producer=Scatter
Unknown Tensor shape: 6777 producer=Scatter
Unknown Tensor shape: 6805 producer=Scatter
Unknown Tensor shape: 6833 producer=Scatter
Unknown Tensor shape: 6836 producer=Conv
Unknown Tensor shape: 6837 producer=Relu
Unknown Tensor shape: 6840 producer=Conv
[...]

PR #629, opened 2019-09-03
Implement Scatter

take-cheeze:
Rel: https://github.com/pfnet-research/chainer-compiler/issues/612

### DIFF
--- a/runtime/ops/indexing.cc
+++ b/runtime/ops/indexing.cc
@@ -191,25 +191,19 @@ chainerx::Array GatherOp::RunImpl(ChxVMState* st, const chainerx::Array& data, c
     return data.Take(indices, axis);
 }
 
-// TODO(take-cheeze): Move to ChainerX
 chainerx::Array ScatterOp::RunImpl(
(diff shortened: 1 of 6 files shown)


runtime/ops/indexing.cc
```diff
@@ -190,6 +191,38 @@ chainerx::Array GatherOp::RunImpl(ChxVMState* st, const chainerx::Array& data, c
     return data.Take(indices, axis);
 }
 
+// TODO(take-cheeze): Move to ChainerX
+chainerx::Array ScatterOp::RunImpl(
+        ChxVMState* st, const chainerx::Array& data_, const chainerx::Array& indices_, const chainerx::Array& updates_) {
+    const int64_t axis = this->axis < 0 ? data_.shape().size() - this->axis : this->axis;
```
shinh: Sorry, now I remembered what we have. I think you can use `BlitArray`. It should be as simple as like
```
slices = make_slice(indices_);
BlitArray(data_, updates_.At(slices));
```

PR #630, opened 2019-09-03
Speedup large tests

shinh:


shinh:
The reduction was rather small, but I'd like to merge this to reduce the size of `out` in my repo. WDYT @ir5 ?

PR #637, opened 2019-09-04
Run tests for expanded dynamic quantize linear

shinh:


shinh:
https://github.com/pfnet-research/chainer-compiler/pull/638

PR #643, opened 2019-09-04
Merge Add/Sum nodes to Sum for PyTorch ONNX

take-cheeze:


take-cheeze:
Seems simplifier works after merging so just extra compilation cost increased...
Memory efficiency is more important in recomputation so closing.

shinh:
Oh, I thnink it's actually a great idea to have this backends other than chxvm.

take-cheeze:
OK. I'll add the config!

PR #665, opened 2019-09-10
Add standard ONNX sequence ops

shinh:
Proposed in https://github.com/onnx/onnx/pull/2249

TODO(hamaji): Remove our own definitions.


shinh:
Removal was done in https://github.com/pfnet-research/chainer-compiler/pull/769 (except `ChainerSequencePop`)

PR #668, opened 2019-09-10
test: Enable subscript assignment test

Rishav1:
Enabling test that uses `ChainerSequenceUpdate` for subscript assignment. Adding a test for `ChainerSetItem` which isn't working properly (Could be an issue in parameter order).


shinh:
Thanks! I'll look into the issue

PR #674, opened 2019-09-10
fix: veval_aug_assign bug fix

Rishav1:
This bug fix brings down mismatch in [fpn.py](https://github.com/pfnet-research/chainer-compiler/blob/master/testcases/elichika_tests/chainercv_model/fpn/fpn.py#L88) to an acceptable failure `Mismatch: 443 / 1605632 (0.0275904%)`.


Rishav1:
Closing PR as this is not the correct fix.

PR #691, opened 2019-09-18
Run tests with dldt and tensorrt on CI

shinh:


shinh:
/test

(this should fail until we merge https://github.com/pfnet-research/chainer-compiler/pull/690)

PR #697, opened 2019-09-19
Support SNPE DLC quantize

take-cheeze:


take-cheeze:
TODO: fix https://developer.qualcomm.com/forum/qdn-forums/software/qualcomm-neural-processing-sdk/36055

take-cheeze:
The library loading problem was due to some python configuration so unrelated.
Some are working so merging

PR #704, opened 2019-09-20
Run tests for type inference on CI

shinh:


chainer_compiler/elichika/typing/types.py
```diff
@@ -349,14 +349,14 @@ def TyChainerVariable(dtype, ndim=None, shape=None):
 
 # ------------------------ TypeChecker internal types --------------------------
 
-counter = 0
+var_counter = 0
```
momohatt: FYI: Actually we don't really need this counter variable to distinguish different TyVars and I'm just keeping it for `show()`. If you think it should be removed, please feel free to do so.

momohatt:
I forgot to mention that we need to change these 2 lines too
https://github.com/pfnet-research/chainer-compiler/blob/f2d62b50d6d608699fdd03c6d6c50829b9d6fd15/chainer_compiler/elichika/testtools/type_checker_tools.py#L97-L98

chainer_compiler/elichika/typing/types.py
shinh: Yeah, I realized this is only for debugging. I personally like these kind of features so I'd like to keep this at least for a while. Let's remove this if it turns out this isn't not very useful for debugging.

shinh:
As for two lines in `type_checker_tools.py`, I modified it locally to update the testcases but decided not to update them in this PR. I think we can/want to introduce a script which re-generates all testcases. Let me see.

PR #705, opened 2019-09-24
Fix paths for module import

shinh:


shinh:
Thanks for the extremely quick response!

I misunderstood what you meant in https://github.com/pfnet-research/chainer-compiler/pull/704#issuecomment-533813306 and I've just realized I needed to change this to append _test suffixes as you said.

PR #708, opened 2019-09-24
Add a script to rebase tests for type inference

shinh:


shinh:
I'd like to introduce a script to update test expectations easily. The expected usage is documented in

https://github.com/pfnet-research/chainer-compiler/pull/708/files#diff-2d5bd2d9f56bf77e8c188af7f315391bR3

WDYT?

momohatt:
Thanks for the great work!!
I feel that the current rebase script generates too much debug prints, so it might be better to turn this `is_debug` off in rebasing?
https://github.com/pfnet-research/chainer-compiler/pull/708/files#diff-ebd6abfcaa5666a8519d2653e36aea67R96

shinh:
Done, thanks for your suggestion!

PR #721, opened 2019-09-29
Progress with FPNFasterResnet50

Rishav1:
Includes support for some numpy operations and chainer functions. Also corrected behaviour to update array when it's subscript is updated. FasterResnet almost compiles with the exception of [anchors](https://github.com/pfnet-research/chainer-compiler/blob/master/testcases/elichika_tests/chainercv_model/fpn/rpn.py#L79) method and [decode](https://github.com/pfnet-research/chainer-compiler/blob/master/testcases/elichika_tests/chainercv_model/fpn/rpn.py#L109) method(requires support for complex index assignment).


Rishav1:
@shinh I don't understand the error from the logs. Could you help me in understanding what test is breaking due to my change? I tried to figure it out this weekend but I couldn't find any clue.

take-cheeze:
Hmm, seems result of chainer and chainer-compiler runtime dfiffers in EspNet.

Rishav1:
I managed to reduce the number of errors from 6 to 2. Now the output graphs are identical except the names, but surprisingly the test gradients don't match.  I've attached the generated `out/elichika_model_EspNet_Decoder_backprop` folder after my changes to elichika -> [elichika_model_EspNet_Decoder_backprop.tar.gz](https://github.com/pfnet-research/chainer-compiler/files/3748113/elichika_model_EspNet_Decoder_backprop.tar.gz)

The test
```
$ build/tools/run_onnx --test out/elichika_model_EspNet_Decoder_backprop --compiler_log --skip_runtime_type_check --backprop
```
and 
```
$ build/tools/run_onnx --test out/elichika_model_EspNet_Decoder_backprop --compiler_log --skip_runtime_type_check --backprop_two_phase
```
are the only two failures I am unable to debug. The dumps of attached model and the one generated by 30c2a1f5 are identical except the names.

shinh:
Sorry, this is a bug in the backend. I think
```
Gradient not supported: ChainerSequenceUpdate
```
is the culprit (probably I should change this to CHECK). I'll send a fix soon.

shinh:
I think this fixes the issue I mentioned: https://github.com/pfnet-research/chainer-compiler/pull/775 I confirmed it actually fixed "elichika_model_EspNet_Decoder_backprop.tar.gz". Thanks a lot for giving us the repro!

shinh:
Merging since this was verified in https://github.com/pfnet-research/chainer-compiler/pull/779. Thanks a lot for this work!

PR #729, opened 2019-10-03
Make clang-format binary configurable

take-cheeze:
And support menoh files too

### DIFF
--- a/scripts/run-clang-format.sh
+++ b/scripts/run-clang-format.sh
@@ -4,5 +4,5 @@ set -eu
 
 cd $(git rev-parse --show-toplevel)
 
-git ls-files | grep -e '\.cc$\|\.h$' | xargs -P4 clang-format -i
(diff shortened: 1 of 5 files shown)


scripts/run-clang-format.sh
```diff
@@ -4,5 +4,5 @@ set -eu
 
 cd $(git rev-parse --show-toplevel)
 
-git ls-files | grep -e '\.cc$\|\.h$' | xargs -P4 clang-format -i
+git ls-files | grep -e '\.cc$\|\.h$|\.cpp$|\.hpp$' | xargs -P4 ${CLANG_FORMAT_BIN:-clang-format} -i
```
shinh: Could you remove `grep` by `git ls-files -- '**/*.h' '**/*.cc' '**/*.hpp' '**/*.cpp'`?

PR #753, opened 2019-10-10
Add test cases that targets older ONNX opset

take-cheeze:
Adds following options to runtests.py:
- `--target_opsets` : comma separated list of target opsets
- `--only_opset_targetable` : runs test cases with `opset_version`


compiler/simplifier.cc
```diff
@@ -720,6 +720,34 @@ bool ReplaceClip(Graph* graph, Node* node) {
...
+bool ReplaceUpsample(Graph* graph, Node* node) {
+    if (node->inputs().size() == 2) {
+        return false;
+    }
+
+    // Upsample-7 to Upsample-9
+    CHECK(node->scales().size() > 0);
+    GraphBuilder gb(graph, "SimplifyUpsample", node->output(0));
+    Value* scales = gb.Const(ArrayBuilder({static_cast<int64_t>(node->scales().size())}).WithData<float>(node->scales()).Build());
+    gb.MOp(Node::kUpsample, {node->input(0), scales}, node->outputs())->set_mode(node->mode());
```
shinh: Probably better to upgrade to Resize since Upsample is deprecated?

PR #760, opened 2019-10-11
fix undefined symbol errors on macOS

msakai:
This fixes errors like following:
```
[100%] Linking CXX shared module run_onnx_core.so
Undefined symbols for architecture x86_64:
  "_PyBaseObject_Type", referenced from:
      pybind11::detail::make_object_base_type(_typeobject*) in run_onnx_core.cc.o
```

I'm using `Apple LLVM version 10.0.0 (clang-1000.11.45.5)` on macOS Mojave.


msakai:
Oh, thanks.
#761 works for me and I think it's more desirable fix.

shinh:
Thanks for the check!

PR #768, opened 2019-10-15
Fixes to runtime/compiler configurations

take-cheeze:
- [x] Split configurations to `{ compiler: {...}, runtime: {...} }`
- [x] Run menoh API test with some configuration example


menoh/CMakeLists.txt
```diff
@@ -59,6 +60,40 @@ if(OpenCV_FOUND)
...
+      SHOW_PROGRESS
+      )
+    # file(
+    #   DOWNLOAD "https://preferredjp.box.com/shared/static/x4k1pya1w1vuvpfk77pp6ma8oaye8syb.onnx"
+    #   "${CMAKE_CURRENT_BINARY_DIR}/resnet50.onnx"
+    #   EXPECTED_HASH SHA256=42916ba7e64f211e1c67aa146e147a01ecd40eaf6b87426dbcd1b7fd9ac8d5d7
+    #   SHOW_PROGRESS
+    #   )
+    file(
+      DOWNLOAD "https://preferredjp.box.com/shared/static/o2xip23e3f0knwc5ve78oderuglkf2wt.onnx"
```
shinh: It's OK for now, but personally, I don't like VGG for testing since it's too large and too simple. ResNet50 is better but I feel it's still too large. I want this test not run by default at least. I think we can switch to shufflenet once https://github.com/onnx/onnx/pull/2328 is merged (currently, chainer-compiler thinks shufflenet is dynamic shape due to the broken shape inference)

menoh/menoh_example.cpp
```diff
@@ -103,27 +103,29 @@ int main(int argc, char** argv) {
...
     vpt_builder.add_input_profile(conv1_1_in_name, menoh::dtype_t::float_, {batch_size, channel_num, height, width});
-    vpt_builder.add_output_name(fc6_out_name);
+    // vpt_builder.add_output_name(fc6_out_name);
     vpt_builder.add_output_name(softmax_out_name);
 
     std::cout << "Build VPT..." << std::endl;
     // Build variable_profile_table and get variable dims (if needed)
     auto vpt = vpt_builder.build_variable_profile_table(model_data);
+    /*
     auto fc6_dims = vpt.get_variable_profile(fc6_out_name).dims;
```
shinh: Ah, I realized you've removed this. It seems shufflenet already runs thanks to this change. However, it seems label mismatches so we should find a right label file to use shufflenet...

take-cheeze: Seems the label mismatch is caused by preprocess mistakes.
I'll check it

Issue #747, opened 2019-10-08
Honor opset_import in ModelProto


shinh:
Currently, we are ignoring `opset_import` when we run shape inference. Currently, `Graph` and `Node` do not have an access to `Model` nor opset_import, so we need to either

1. add `Graph::model_` and `Node::model_` (or `Node::graph_`)
2. introduce `ModelScope` or something which sets `Model::current_model()` and use it in passes.cc

I'd prefer the latter approach.

Anyway, chainer_compiler_core.cc should be also updated so deprecate `LoadGraph` and introduce `LoadModel` instead.

shinh:
https://github.com/pfnet-research/chainer-compiler/pull/748/commits/b5574aea26b7ec7b5798e26f005d813f258bf69f could be reconsidered after fixing this issue.

take-cheeze:
Merged https://github.com/pfnet-research/chainer-compiler/pull/770 so closing

PR #770, opened 2019-10-16
Add opset imports to Node and Graph

take-cheeze:
Rel: https://github.com/pfnet-research/chainer-compiler/issues/747

- Support ONNX Sequence inference
- Pass opset imports explicitly when constructing graph
- Check inputs/outputs count and required attribute when serializing to protobuf
- Merge https://github.com/pfnet-research/chainer-compiler/pull/750 (closes #750 )
- Constant propagate `Neg` op


compiler/node.h
```diff
@@ -71,12 +76,21 @@ class Node : public NodeBase {
...
+
+    bool ValidateWithSchema(const OpsetList& opsets = {}, std::string* message = nullptr) const;
+
 private:
     std::vector<Value*> inputs_;
     std::vector<Value*> outputs_;
     std::string name_;
     std::string domain_;
     std::string doc_string_;
+    OpsetList opset_import_;
```
shinh: I want all non-ephemeral states of these classes to be serialized in ONNX so that we can stop and resume graph transformation at any stage. Could you introduce custom attributes to serialize this info? Probably:

```
CHAINER_COMPILER_GLOBAL_ATTRS = attr_sets(chainer_order=-1, chainer_fusion_group=0, chainer_onnx_domain=[str], chainer_onnx_version=[int])
```

? (I found a typo in `CHAINER_COMPILERX_GLOBAL_ATTRS`)

take-cheeze: Making it to attribute sounds nice :)

chainer_compiler/elichika/functions_builtin.py
```diff
@@ -306,13 +306,22 @@ def __init__(self):
...
         parser = self.parse_args(onnx_graph, node)
 
-        onnx_graph.add_node(
-            "Clip",
-            [parser.get('x')],
-            node.outputs,
-            str(node.lineprop),
-            min=parser.get('x_min'),
-            max=parser.get('x_max'))
+        if onnx.defs.onnx_opset_version() >= 11:
```
shinh: Ah, I didn't know this function :)

PR #781, opened 2019-10-24
Add libcuda.so.1 link in chainer-compiler build on docker image

take-cheeze:


shinh:
What is the motivation of this change? I guess we should drop `/usr/local/cuda/lib64/stubs` from `LD_LIBRARY_PATH` and use `/usr/lib/x86_64-linux-gnu/libcuda.so.1` on GPU?

take-cheeze:
Seems I've missed this comment:
https://github.com/pfnet-research/chainer-compiler/blob/978f4915e9dfd0ed32a40b30b096fb3dbc34c6ba/docker/Dockerfile#L224-L225
Never mind for this! 🙇

PR #805, opened 2020-02-29
Refine type inference for arithmetic tensor operation

momohatt:
This PR includes:
* Refine inference of arithmetic tensor operation (e.g. add, sub, mul, etc.)
  * Choose "smaller" expression when the math expression for shapes collide
* Remove `dtype=` and `shape=` from the pretty printing of `TyTensor`


chainer_compiler/elichika/typing/ext/common.py
```diff
@@ -0,0 +1,78 @@
...
+            shape = self.infer_return_shape(x_shape, y_shape)
+        else:
+            shape = self.infer_return_shape(y_shape, x_shape)
+        return TyTensor(self.kind, dtype, shape)
+
+    def infer_return_shape(self, x_shape, y_shape):
+        ret_shape = [None for _ in x_shape]
+        for i in range(len(x_shape) - len(y_shape)):
+            ret_shape[i] = copy_ShapeElem(x_shape[i])
+        for i in range(1, len(y_shape) + 1):
```
shinh: Probably, `reversed(range(len(y_shape)))` (and s/-i/i/g) would be easier to read?

momohatt: Thanks! I think it's equivalent to `range(len(x_shape) - len(y_shape), len(x_shape))` and s/`y_shape[i]`/`y_shape[i - len(x_shape) + len(y_shape)]`/g .

shinh: Ah, I read code wrongly. I thought the size of `y_shape` is equal to `x_shape`/`ret_shape`. Then, I think the original code looks slightly better to me, but the current code would also make sense. Sorry for the back-and-forth. It's up to you.

momohatt: I prefer the original version, too. I'll revert the commit. Thanks!

PR #810, opened 2020-03-04
Improve type inference for PyTorch implementation of VGG2L in EspNet

momohatt:


chainer_compiler/elichika/typing/types.py
```diff
@@ -583,6 +590,16 @@ def tyobj_to_dtype(ty):
     return np.dtype(str(NumKind(ty.kind)))
 
 
+def dtype_to_tyobj(dtype):
```
shinh: It seemed this function is not called from the unittests so I don't understand the expected value of `dtype`, but I guess this tries to handle both `np.int32` and `np.dtype('float32')`? They are difficult to me

```
>>> type(np.int32)
<class 'type'>
>>> type(np.dtype('int32'))
<class 'numpy.dtype'>
>>> np.int32 == np.dtype('int32')
True
```

but I guess what you want is something like

```
# https://docs.scipy.org/doc/numpy/reference/generated/numpy.dtype.kind.html
def dtype_to_kind(dtype):
    dtype = np.dtype(dtype)
    if dtype.kind == 'b':
        return 'bool'
    if dtype.kind in 'iu':
        return 'int'
    if dtype.kind == 'f':
        return 'float'
    assert False
```

?

momohatt: tbh I don't remember why I put the first `if` statement like that, and this indeed seems strange... This `dtype_to_tyobj` gets called in the inference of `numpy.tolist()` and `torch.Tensor.tolist()`, so I guess we can safely assume that the argument `dtype` will always be of type `numpy.dtype`. I'll fix it.

PR #814, opened 2020-03-16
Add type guard for optional types

momohatt:
This PR implements basic type guard (like the one in TypeScript) for optional types.
It narrows down the type of the variable being tested in the `if` statement as follows.
```
# x : optional(int)
if x is None:
    # x : NoneType
else:
    # x : int
```
```
# x : NoneType
if x is None:
    # x : NoneType
else:
    # x : a (fresh TyVar)
```
```
# x : int
if x is None:
    # x : NoneType
else:
    # x : int
```


shinh:
lgtm!

PR #818, opened 2020-03-23
Fix type inference for dictionaries

momohatt:
The inference results of a dictionary `{'foo': 1, 'bar': None}` has been `{string : int}` whereas it is expected to be `{string : optional(int)}`. This PR fixes such wrong behavior.


shinh:
LGTM, thanks!!

PR #820, opened 2020-03-25
Infer type for assignment where lhs is Subscript

momohatt:


momohatt:
@hamaji Do you have any idea why the test is failing? I couldn't figure it out myself.

take-cheeze:
Strange build error in macOS, so restarting the job.

take-cheeze:
Got it. It's caused by cmake 3.17 in macOS's brew.
Could you apply this?:
```
 scripts/elichika_tests.py | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/scripts/elichika_tests.py b/scripts/elichika_tests.py
index 661cbd8a..f3638bca 100755
--- a/scripts/elichika_tests.py
+++ b/scripts/elichika_tests.py
@@ -132,6 +132,8 @@ def print_test_generators(dirname):
         tests.append(
             os.path.join('testcases/elichika_tests', gen.dirname,
                          gen.filename + '.py'))
+    base_dir = os.path.dirname(os.path.dirname(__file__))
+    tests = [os.path.join(base_dir, t) for t in tests]
     print(';'.join(tests))
 
 ```

take-cheeze:
#821 has the same error so maybe landing #822 first is simpler

momohatt:
Thanks for the fix!

shinh:
LGTM!

